PR #4, opened 2020-08-06
Don't try to load module if there is none

Hipska:
This raises a warning otherwise.


piRGoif:
New PR replaces this one : #5

PR #5, opened 2020-08-19
Don't try to load module if there is none

Hipska:
This would generate a warning every run. Now I made it an info log message.

Also fixed a php warning of glob returning false instead of empty array in [some cases](https://3v4l.org/qZ4hC).

This will fix #4, I had problems updating that branch.

### DIFF
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -265,16 +265,17 @@ abstract class Collector
 	protected function GetVersionFromModuleFile()
 	{
 		$aFiles = glob(APPROOT.'collectors/module.*.php');
-		$sModuleFile = null;
-		$sModuleFile = reset($aFiles);
-		if ($sModuleFile == null)
+		if (!$aFiles)
 		{
 			// No module found, use a default value...
 			$this->sVersion = '1.0.0';
+			Utils::Log(LOG_INFO, "No module file found, using default version value.");
+			return;
 		}
 		
 		try
 		{
+			$sModuleFile = reset($aFiles);
 			$sModuleFileContents = file_get_contents($sModuleFile);
 			$sModuleFileContents = str_replace(array('<?php', '?>'), '', $sModuleFileContents);
 			$sModuleFileContents = str_replace('SetupWebPage::AddModule(', '$this->InitVersionCallback(', $sModuleFileContents);


piRGoif:
Ok, I'm closing #4 !

core/collector.class.inc.php
```diff
@@ -265,16 +265,17 @@ protected function GetCharset()
 	protected function GetVersionFromModuleFile()
 	{
 		$aFiles = glob(APPROOT.'collectors/module.*.php');
-		$sModuleFile = null;
-		$sModuleFile = reset($aFiles);
-		if ($sModuleFile == null)
+		if (!$aFiles)
```
dflaven, line 268: Should we write this if ($aFiles === false) ?

Hipska: I think it can also be an empty array in some cases..

Hipska: You agree?

Hipska: @dflaven?

odain-cbd: yes you are right regarding glob()

 * @return array|false an array containing the matched files/directories, an empty array

core/collector.class.inc.php
```diff
@@ -265,16 +265,17 @@ protected function GetCharset()
...
 		{
 			// No module found, use a default value...
 			$this->sVersion = '1.0.0';
+			Utils::Log(LOG_INFO, "No module file found, using default version value.");
+			return;
```
odain-cbd, line 273: logging is great but we cannot afford to leave the return. lot of custom collector developpers forget to provide a module file. their own collector would not work anymore after migrating collector-base.

Hipska: This PR exactly aims to fix this situation. For those collectors, this throws a php warning every time, after merge, you will only get this INFO log message.

odain-cbd: logging is good to me. only 'return' bothers me. 

when no module file provided, collector will stop its usual processing (earling failure). which can be considered as a regression for our community.

Hipska: It stops the method `GetVersionFromModuleFile` which is exactly what you want, because there is no module file. Line 271 sets the version already. There simply isn't anything else to do in this method.

odain-cbd: ok agreed.

odain-cbd:
thanks for your contribution Hispka. code is better that way.

Hipska:
Erm, why close it and not merge?

odain-cbd:
sorry I am not so fond of github UI. i had to reopen and really merge

Hipska:
Hi, could you please also add the `hacktoberfest` tag to this repo? https://github.com/digitalocean/hacktoberfest/pull/596

Molkobain:
Hello @Hipska , can you argue on why we should tag the repo and how it benefits the repo? This has nothing to do with iTop right? (I didn't read the post)

Hipska:
It gives rewards to people that contribute open source during October. But only on repo's that have the tag, this is a recent change to fight spam PR's.
The benefit might be that others will find it and may contribute as well.

Molkobain:
Okay thanks for the explanation, I'll check with the rest of the team right today.

Not sure it will bring much light on iTop but it won't do any harm either! And above all, if it can be a little token of our appreciation for all your work on the iTop ecosystem, I'm more than ok with it 😊

Molkobain:
The topic was just added to this repo and iTop's. Thanks for your contributions @Hipska ! 😊

PR #6, opened 2020-08-19
Fix other request errors displayed as successful

Hipska:
The `Collector::Synchronize` method does not check if synchro has succeeded. 

In my case, php was throwing a memory limit error and the data collector said it was succeeded, but in the iTop console I saw it was still running (meaning it was broken down).

### DIFF
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -674,6 +674,12 @@ abstract class Collector
 				}
 			}
 		}
+		else
+		{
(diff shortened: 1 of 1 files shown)


dflaven (approved):

Hipska:
@odain-cbd Why is it closed? Is my fix not wanted? Or is the fix incorrect?

odain-cbd:
sorry i will merge it and close it.

Hipska:
😂 If you merge, it's automatically closed. No need to do that manually 😉. But thanks, it took me a while to figure this out.

PR #7, opened 2020-08-27
N°5139 - Let data that is null also be null at the other side

Hipska:
Data that is `null` is uploaded as an empty string, which caused problems on iTop side when this is an integer or decimal field. This PR aims to solve this, by putting `<NULL>` in the csv, which is recognised on iTop side as a null value.


core/collector.class.inc.php
```diff
@@ -14,6 +14,8 @@
 //   You should have received a copy of the GNU Affero General Public License
 //   along with this application. If not, see <http://www.gnu.org/licenses/>
 
+define('NULL_VALUE', '<NULL>');
+
```
odain-cbd, line 18: Hi Hipska,

could you please detail the functionnal goal of this PR? what is the behaviour with current code ? what it will be after?

BR
Olivier

Hipska: This is just the definition of a constant. Similar as in iTop. If you are referring tho the whole PR, please comment at the correct place. The initial post already explains the goal.

Hipska:
@odain-cbd See additional context in https://github.com/Combodo/iTop/pull/152

The problems on iTop side are already solved or didn't need a fix, but now this PR aims to fix the collector side as well.

Hipska:
Example data:
```php
$aData[] = array(
    'primary_key' => 1,
    'name' => 'Foo',
    'number' => 24,
);
$aData[] = array(
    'primary_key' => 2,
    'name' => 'Bar',
    'number' => null,
);
```

Before csv:
```csv
primary_key,name,number
1,Foo,24
2,Bar,
```
After csv:
```csv
primary_key,name,number
1,Foo,24
2,Bar,<NULL>
```

The before csv caused a SQL error at iTop side for the integer field.

Hipska:
Hi @odain-cbd, do you have an update?

Hipska:
@piRGoif maybe you can have a look?

piRGoif:
> @piRGoif maybe you can have a look?

Sorry, I don't know enough about collectors to have an opinion :/

Hipska:
Thanks, could you give @odain-cbd a ping?

Hipska:
Can someone give any response please?

odain-cbd:
Sorry for the delay. I have been busy somewhere else.

from what I remember this PR could not be taken as default behaviour. let me search for the exact explanation. 

as a compromise i will also check if it could be a behaviour triggered on a specific collector configuration (feature set).

Regards
Olivier

odain-cbd:
Here is data synchro doc:
https://wiki.combodo.com/doku.php?id=2_7_0:install:230_to_240_migration_notes&s[]=data&s[]=synchro&s[]=null#data_synchro

" a NULL value has the meaning Ignore this column so don't update the destination object.
an empty string means empty the destination object field (force a reset of a value),"

Today in 2.7.4 resetting fields is possible for string and dates. but not for enum/int and decimal yet. there is a remaining bug N°776 that will be addressed.

your PR does not seem to follow that logic, does it?

BR
Olivier

Hipska:
Note that the link is private, but the article you referring to is purely for date fields. The problem actually is when dealing with integer fields.

Hipska:
@odain-cbd ?

Hipska:
Anyone?

odain-cbd:
Dear Hispka,

sorry for my silence. let me reformulate things around this PR. that should make sure we are on the same page. 

FYI I will use my current answer to share with other Combodo colleagues and see if they agree to what I suggest. and also give more chances to your PR to be accepted....

### **Goal:**
when collecting data you may have null values for some fields (like integer). in that case, at synchronization step the object cannot be loaded as there is a mysql error. 

your PR aims at passing SQL NULL value during datasynchro for any field/any type of field (decimal/string/enum/date/...whatever) when collected value is null. which leads to **IGNORING** the field value on iTop side. 
**on iTop side the object will be created (respectively updated) leaving its value to either default (respectively previous) value.**

### **Objections**

- you decided to IGNORE the field. another design would have been to choose to RESET the field instead.
- current change impacts any field/any type of field during collecting step. this could be seen as a breaking change. other collector users/designer (Combodo partners/customers/community) may prefer to have an early failure and redesign their own collector to avoid to have null value.

### **My suggestion**
I would suggest to make current PR changes configurable to be able to decide for which fields we want to either IGNORE/RESET/FAIL at collecting step with null values. 

**Data synchronization recall**

- passing NULL (SQL) value means IGNORING a field and make sure synchronization goes on and stores the object creation/update on iTop side. (your previous PR152/bug N°3251)

https://github.com/Combodo/iTop/pull/152

- passing empty values SHOULD mean RESETTING the field during data synchronization
this is already current behaviour (iTop 2.7) for below fields: 
enum, text, password, date, integer, decimal, pourcentage, html, email

but there is a limitation (Planned Bug N°776) for enum, integer, decimal (and others like external keys).

BR
Olivier

Hipska:
Okay, how do you see this option for null treatment configuration?

IMO sending a null to iTop synchro means that I want this value to be inserted into the DB like that. I don't get why anyone would want it to be ignored? If you want that, you would set the `SynchroAttribute::update` to `0` if you want to ignore the values from that field, no?

odain-cbd:
"IMO sending a null to iTop synchro means that I want this value to be inserted into the DB like that. I don't get why anyone would want it to be ignored? If you want that, you would set the SynchroAttribute::update to 0 if you want to ignore the values from that field, no?"

for temporary reasons you can decide to let your collector synchronize partial data. for example if you are in a dev environnement and server on which your collect works is not in final mode. or for temporary reasons in production if same server has live issues and cannot provide all the data.

in dev environment you can be flexible (null values accepted and itop object partially synchronized) at the beginning of your project and when you are close to deploy everything in production you change your conf to something more strict (no null values accepted)....

Hipska:
I still don't think you are following the issue. How could I ever sync a number that also can contain `null` values?

If I understand your comment well, the current behaviour works well for temporary situations or during development. While my situation is problematic during production in real world scenario where the value needs to be `null` not `0`.

Hipska:
Solved conflicts.

@odain-cbd could you please answer my question from September?

odain-cbd:
i understand you need to make it work in production. ignoring null will let collector work with a successfull data synchro. data will be updated except the field to ignore.

this choice may not be the one for other customers that prefer to be notified by an error and see why it happens...

Hipska:
Please tell me how and AttributeDecimal field can be synchronised with a NULL value then?

odain-cbd:
AttributeDecimal  cannot be synchronized with NULL. you can either
- work on collect side to avoid null values
- set SynchroAttribute::update should be set to 0
- use _ignored_attributes in params.distrib.xml for json or mysql collectors
`
         * <mysqlcollector_ignored_attributes type="array">
	 *    <attribute>location_id</attribute>
	 * </mysqlcollector_ignored_attributes>
`
cf https://wiki.combodo.com/doku.php?id=extensions:itop-data-collector-base

`
New in 1.0.6 To take into account the possible variations of the data model, without re-writing a collector each time, it is possible to mark some of the collected attributes as “optional” so that the collector can run even if the corresponding attribute does not exist in the data model. Supply an array of attribute codes to ignore, here. `

odain-cbd:
maybe we could add another conf section to list attributes we can ignore with null. something like
`
<mysqlcollector_ignored_null_attributes type="array">

    location_id
    </mysqlcollector_ignored_null_attributes >
`

odain-cbd:
maybe we could even have a more clever configuration. what about a section to configure how we intend to treat some fields when they are null when collected. 
```
<mysqlcollector_null_attributes type="array">

    <location_id>ignore</location_id>
    <location_id>reset</location_id>
    </mysqlcollector_null_attributes >
```

by default a field with null would fail if not listed (backward compatibility).

PS:
besides i would like to apologize for my delay responses. this is hard for me to keep track on github side and remain focused on my current projects...

odain-cbd:
Dear Hispka,

After our call, I came back to our product owner team. your fix seems nice and is a real workaround to N°776 bug related to enum, integer and decimal fields in iTop.

But we cannot accept your work as is, as it could be seen as a regression for other type of fields. to keep backward compatibility, as discussed together as well, we could change collector behaviour on a subset of fields with an additional configuration in params.distrib.xml:
```
  <mycollector_nullified_attributes type="array">
    <attribute>field1</attribute>
    <attribute>field2</attribute>
    ...
  </mycollector_nullified_attributes >
```

for field1/fields2, any null value will be transformed to '<NULL>' before data synchro like proposed in current pull request.
for others fields, no transformation will be done.

are you ok with this design? if so, please adapt current PR accordingly. on my side i will create a bug on Combodo side to release it soon and will document the collector-base wiki documentation. regarding tests, i will make my best to cover this work as well.

Best regards,
Olivier

Molkobain:
Accepted in technical review

core/collector.class.inc.php
```diff
@@ -137,6 +140,47 @@ public function AttributeIsOptional($sAttCode)
...
+	 * @return boolean True if the attribute can be skipped, false otherwise
+	 */
+	public function AttributeIsNullified($sAttCode) {
+		if ($this->aNullifiedAttributes === null) {
+			$this->aNullifiedAttributes = Utils::GetConfigurationValue(get_class($this)."_nullified_attributes", null);
+			if ($this->aNullifiedAttributes === null) {
+				// Try all lowercase
+				$this->aNullifiedAttributes = Utils::GetConfigurationValue(strtolower(get_class($this))."_nullified_attributes", null);
+			}
+		}
```
Hipska, line 173: @odain-cbd Shouldn't this part be done in the class constructor? Now it will run this code for every field for every row if there aren't any nullified attributes..

odain-cbd: that's right. i had some doubt as well... 

I had the choice between doing like this and code will be executed further times for every field if there is no such configuration BUT only during collecting step. or moving it inside constructor and it will be executed once even for other steps where it does not make sense... 
maybe we could adapt that code to follow your advice (moving it into constructor) and checking which step will be executed (collecting or not).

FYI  method AttributeIsOptional works also like AttributeIsNullified and could be moved inside constructor as well.

Hipska: Okay, I have updated the code.

Hipska:
Hi, Is there any update on this?

PR #8, opened 2020-12-03
Allow login_mode to be set

Hipska:
In the case where the first `allowed_login_types` is for example `saml`, then the calls to the API will be redirected to the SAML identity provider, but this is not supported by the curl calls.

Solution is to put `login_mode=form` as an extra URL parameter. I have made it configurable in case of any other situation that is not using form by default.

### DIFF
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -581,6 +581,7 @@ abstract class Collector
 			$this->DoProcessBeforeSynchro();
 		}
 		
+		$sLoginMode = Utils::GetConfigurationValue('itop_login_mode', 'form');
 		$aFiles = glob(Utils::GetDataFilePath(get_class($this).'-*.csv'));
 		foreach($aFiles as $sDataFile)
 		{
@@ -598,7 +599,7 @@ abstract class Collector
 				'charset' => $this->GetCharset(),
 			);
 			$sUrl = Utils::GetConfigurationValue('itop_url', '').'/synchro/synchro_import.php';
-			$sUrl .= '?login_mode=' . Utils::GetConfigurationValue('itop_login_mode', 'form');
+			if (!empty($sLoginMode)) $sUrl .= '?login_mode=' . $sLoginMode;
     		$iSynchroTimeout = (int)Utils::GetConfigurationValue('itop_synchro_timeout', 600); // timeout in seconds, for a synchro to run	
 
 	    	$aResponseHeaders = null;
@@ -641,7 +642,8 @@ abstract class Collector
 			$aData['max_chunk_size'] = $iMaxChunkSize;
 		}
 		$sUrl = Utils::GetConfigurationValue('itop_url', '').'/synchro/synchro_exec.php';
-		$sUrl .= '?login_mode=' . Utils::GetConfigurationValue('itop_login_mode', 'form');
+		if (!empty($sLoginMode)) $sUrl .= '?login_mode=' . $sLoginMode;
+
 		$iSynchroTimeout = (int)Utils::GetConfigurationValue('itop_synchro_timeout', 600); // timeout in seconds, for a synchro to run
 		
 		$aResponseHeaders = null;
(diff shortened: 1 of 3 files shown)


piRGoif (approved):

piRGoif:
Thanks for this !
As a reference, login_mode parameter is documented here : https://www.itophub.io/wiki/page?id=2_2_0%3Aadmin%3Auser_authentication_options#login_mode_page_parameter

piRGoif:
Hello, 
After a team review we would like to keep this optional : so that by default no parameter would be added to the URL, and only if the config parameter is present then add the parameter in the URL.

Hipska:
That makes sense indeed. I will have a look.

core/collector.class.inc.php
```diff
@@ -598,6 +599,7 @@ public function Synchronize($iMaxChunkSize = 0)
 				'charset' => $this->GetCharset(),
 			);
 			$sUrl = Utils::GetConfigurationValue('itop_url', '').'/synchro/synchro_import.php';
+			if (!empty($sLoginMode)) $sUrl .= '?login_mode=' . $sLoginMode;
```
piRGoif, line 602:
```suggestion
			if (!empty($sLoginMode)) { $sUrl .= '?login_mode=' . $sLoginMode;}
```

core/collector.class.inc.php
```diff
@@ -640,6 +642,8 @@ public function Synchronize($iMaxChunkSize = 0)
 			$aData['max_chunk_size'] = $iMaxChunkSize;
 		}
 		$sUrl = Utils::GetConfigurationValue('itop_url', '').'/synchro/synchro_exec.php';
+		if (!empty($sLoginMode)) $sUrl .= '?login_mode=' . $sLoginMode;
```
piRGoif, line 645:
```suggestion
		if (!empty($sLoginMode)) { $sUrl .= '?login_mode=' . $sLoginMode;}
```

core/restclient.class.inc.php
```diff
@@ -108,8 +108,10 @@ protected static function ExecOperation($aOperation, $sVersion = '1.0')
 		$aData['auth_user'] = Utils::GetConfigurationValue('itop_login', '');
 		$aData['auth_pwd'] = Utils::GetConfigurationValue('itop_password', '');
 		$aData['json_data'] = json_encode($aOperation);
-//print_r($aOperation);
+
 		$sUrl = Utils::GetConfigurationValue('itop_url', '').'/webservices/rest.php?version='.$sVersion;
+		$sLoginMode = Utils::GetConfigurationValue('itop_login_mode', 'form');
+		if (!empty($sLoginMode)) $sUrl .= '&login_mode=' . $sLoginMode;
```
piRGoif, line 114:
```suggestion
		if (!empty($sLoginMode)) { $sUrl .= '&login_mode=' . $sLoginMode;}
```

piRGoif (changes requested):
Thanks for the quick modification !
But I would ask a little form change : add the braces for the if, even they are on one line ?

Hipska:
Does this module follow other coding convention than iTop? As it is allowed to have a single line `if` without braces over there, and it is used a lot in the current codebase.

piRGoif:
Indeed, the [coding standards](https://www.itophub.io/wiki/page?id=latest%3Acustomization%3Acoding_standards) page doesn't tell a word about this, and in the existing codebase, either iTop or iTop modules, there are a lot of such if without braces. Lots of this legacy code was developed using simple text editors, limited formatting rules... 

We tend to change this step by step since some years now, but we can't refactor the whole codebase as this would make the history harder to read (especially the blame function in large and old files like MetaModel or such)
For example our iTop .editorconfig has some rules on enforcing braces, like [this one for PHP](https://github.com/Combodo/iTop/blob/develop/.editorconfig#L347). 

This is a good practice as avoiding them could cause great confusions, like adding a line after the if thinking it would be part of it... and it's not !
Adding braces doesn't cost much, so :)

piRGoif (approved):

piRGoif:
Thanks, I'm merging this PR !

piRGoif:
After some discussions with @dflaven, @etaloc and of course @Hipska we decided to hardcode the login_mode.
See ed04f2d1

piRGoif:
Combodo internal ref : N°3559

PR #9, opened 2021-03-10
N°6727 Fix config dump with special characters

Hipska:
When a config item contains special characters like `&`, which is very common for the LDAP collector, the config dump ignores the value.

My contribution fixes that.

See before (foo) and after (bar) example: https://3v4l.org/AV8kO


piRGoif:
Would you mind create a PHPUnit test to cover this modification ?

piRGoif (approved):

core/parameters.class.inc.php
```diff
@@ -55,19 +55,19 @@ protected function ToXML(DOMNode $oRoot, $data = null, $sNodeName = null)
 		{
 			$data = $this->aData;
 		}
-				
-		if (is_array($data))
```
piRGoif, line 59: Woops sorry, second thought... 
This if has moved, so now if `$data` isn't an array then now we will create an element... Is this change really intended ?

piRGoif (changes requested):

core/parameters.class.inc.php
Hipska: Yes, it's moved to line 69 and 70.

piRGoif: This change the code behavior... So I'm worried about regressions... Can you explain why did you move this line ?

Hipska: Because the element needs to be created also in the case it's not an array.

piRGoif: Oh ok, I didn't noticed the modifications in the else block ! Before we had a createElement() call, now this is always done above and we're adding a textnode as a child. Sorry.

In theory ok for me so I'm closing this conversation. But... this is a supplementary reason to add a test. This PR won't be accepted without it.

Hipska: Great, still need to look into that. But you know, time for this is limited.

piRGoif: Yes I understand.
If I ever find some time to write the test myself, did you checked the "allow maintainers" checkbox on the PR ?

Hipska: Yes I believe I did

Molkobain:
Hello Thomas, we are planning a code freeze of collectors for Feb 24th. If you want the PR to be merged by then, please provide the unit test before.

Molkobain:
@Hipska should we no longer consider this PR?

Hipska:
Is the bug fixed already?

odain-cbd:
Dear Thomas,

Sorry if i came so late. I just discovered current PR and it seems we encountered similar issues. cf branch feature/6120-dumpconfig-fix. could you please have a look?

main differences:

- I added a test with a XML example to document the fix
- my code is simpler but may not cover some of the usecases you encountered

Do you mind if i help integrating your PR by merging your work with mine? 
could you please give us a XML example that may explain the difference between your code and mine? that will help understand and cover this part of the code....

BR 
Olivier

odain-cbd:
Yes it is a good idea to merge my branch into current PR. for that purpose I created a PR to merge it in your fork.

BR

Hipska:
Thanks, I merged master into this branch to reduce the clutter, but there is still a conflict.

odain-cbd:
SaaS team work has been merged in current fork. a test covers it...

piRGoif:
Thanks to both of you, I'll try to take a look tomorrow, before the next PR review on tuesday.

Molkobain (approved):

Hipska:
I'm not really convinced of the provide test case,

Wouldn't this be more appropriate?
```php
		$oParams = new \Parameters($sXmlPath);
		$oParams->SaveToFile($this->sTempParamsFile);

		$oNewParams = new \Parameters($this->sTempParamsFile);

		$this->assertEquals($oParams->Get('broken_param'), $oNewParams->Get('broken_param'));
```

(as a proof of concept, as the dump is not actually using this save function)

piRGoif:
Sorry running late... I have a deadline due to this evening. Will review as soon as possible.

core/parameters.class.inc.php
```diff
@@ -55,13 +55,14 @@ protected function ToXML(DOMNode $oRoot, $data = null, $sNodeName = null)
 			$data = $this->aData;
 		}
 
+		if ($oRoot instanceof DOMDocument) {
+			$oNode = $oRoot->createElement($sNodeName);
+		} else {
+			$oNode = $oRoot->ownerDocument->createElement($sNodeName);
+		}
+		$oRoot->appendChild($oNode);
+
```
piRGoif, line 64: Why was this block moved outside of the `if (is_array($data))` test ?

piRGoif (commented):
Sorry it took me so long to do this review :/ 

Many thanks for the detailed example in the description, very clear !
Just have a question on the createElement() which was moved ?

core/parameters.class.inc.php
Hipska: Because it is also needed for the `else`

piRGoif:
> I'm not really convinced of the provide test case,

I agree adding the assertEquals on the broken_param node would make the test clearer ! 
Also using directly the Parameters method is more atomic than working with the Utils class...

@odain-cbd what do you think ?

piRGoif (approved):
Approving this PR but... see in the conversation tab about the test that can be improved

odain-cbd:
I added Hispka testcase in feature/6120-dumpconfig-fix branch. but i still think both tests are valuable. 

Mine makes sure we read properly XML file. it is covering the dumptask option in CLI. it  this is something we rely on for another LDAP Configuration extension from UI (SaaS). 
Yours is covering persistency in both I/O sides (save and read). it is easier to read.  more atomic as well (agree with Pierre).

Hipska:
The only thing I'm having a problem with is that your test requires the xml to be exactly the same. While an xml with elements in a different order would result in the same configuration as expected, but the test would fail. Both xml and php arrays don't have a guaranteed fixed ordering of elements.

I would be fine if the test would say for example it should contain `<broken_param>(&amp;(objectClass=person)(mail=*))</broken_param>`.

piRGoif:
Hello,
We just did a review with Olivier ! 
We are proposing adding a new test/ParametersTest file to test the modified method more atomically.
Here is its content : 

```php
<?php
/*
 * @copyright   Copyright (C) 2010-2023 Combodo SARL
 * @license     http://opensource.org/licenses/AGPL-3.0
 */

namespace UnitTestFiles\Test;

use Parameters;
use PHPUnit\Framework\TestCase;

@define('APPROOT', dirname(__FILE__, 2).'/');

require_once(APPROOT.'core/parameters.class.inc.php');

class ParametersTest extends TestCase {
	/**
	 * @dataProvider ToXMLProvider
	 */
	public function testToXML(array $aData, string $sExpectedDump) {
		$oParameters = new Parameters();
		$this->SetNonPublicProperty($oParameters, 'aData', $aData);

		$sDumpParameters = $oParameters->Dump();

		$this->assertStringContainsString($sExpectedDump, $sDumpParameters);
	}

	public function ToXMLProvider() {
		return [
			'Parameter with &amp;' => [
				'aData' => ['escaped_param' => '(&(objectClass=person)(mail=*))'],
				'sExpectedDump' => '<escaped_param>(&amp;(objectClass=person)(mail=*))</escaped_param>',
			],
			'Parameter with array' => [
				'aData' => ['paramroot' => ['param1' => 'param1val', 'param2' => 'param2val']],
				'sExpectedDump' => "<paramroot>\n    <param1>param1val</param1>\n    <param2>param2val</param2>\n  </paramroot>",
			],
		];
	}

	public function SetNonPublicProperty(object $oObject, string $sProperty, $value)
	{
		$oProperty = $this->GetProperty(get_class($oObject), $sProperty);
		$oProperty->setValue($oObject, $value);
	}
	private function GetProperty(string $sClass, string $sProperty): \ReflectionProperty
	{
		$class = new \ReflectionClass($sClass);
		$property = $class->getProperty($sProperty);
		$property->setAccessible(true);

		return $property;
	}
}
```

I don't have the rights to push on your fork, can you add it for us ? 

About testing node order, we are proposing to add it outside of this PR, as here we want to test the modification done especially, and this PR was opened a long time ago.

Hipska:
@piRGoif I’m off today, but you could add a PR to our branch on our repo fork. I will look into giving you rights to our fork later…

piRGoif:
> @piRGoif I’m off today, but you could add a PR to our branch on our repo fork. I will look into giving you rights to our fork later…

Done : https://github.com/Super-Visions/itop-data-collector-base/pull/3

Hipska:
@piRGoif The second part of this test is still relevant to this comment.

> The only thing I'm having a problem with is that your test requires the xml to be exactly the same. While an xml with elements in a different order would result in the same configuration as expected, but the test would fail. Both xml and php arrays don't have a guaranteed fixed ordering of elements.

I think the best test would be a dump and read of the xml which should result in the same config..

piRGoif:
> The second part of this test is still relevant to this comment.

We may have this conversation in the corresponding PR ? https://github.com/Super-Visions/itop-data-collector-base/pull/3

Hipska:
@piRGoif My comment didn't have to do with the test you added in that PR, so maybe still continue here?

piRGoif:
> @piRGoif My comment didn't have to do with the test you added in that PR, so maybe still continue here?

Sorry I don't understand. To which test are you refering to ? I think this is the one Olivier made ? UtilsTest::testDumpConfig ? But the one I added does what you are asking for ?

Hipska:
Yes, as it was a reply to @odain-cbd and your test wasn't created back then..

piRGoif:
Ok ! We have 2 tests, one global and one on a deeper level, I'm planning this PR to be reviewed next month.

piRGoif:
Technical review : OK

piRGoif:
Functional review : OK for next collector version

PR #10, opened 2021-03-16
Use php included JSON pretty print

Hipska:
Note that I added deprecation warning to next version release as this will bump minimal required php version to 5.4.0

### DIFF
--- a/core/utils.class.inc.php
+++ b/core/utils.class.inc.php
@@ -439,7 +439,7 @@ class Utils
 
 	/**
 	 * Pretty print a JSON formatted string. Copied/pasted from http://stackoverflow.com/questions/6054033/pretty-printing-json-with-php
-	 * @deprecated 1.3.0 use `json_encode($json, JSON_PRETTY_PRINT);` instead
+	 * @deprecated 1.3.0 use `json_encode($value, JSON_PRETTY_PRINT);` instead
 	 * @param string $json A JSON formatted object definition
 	 * @return string The nicely formatted JSON definition
 	 */
--- a/core/utils.class.inc.php
+++ b/core/utils.class.inc.php
@@ -439,7 +439,7 @@ class Utils
 
 	/**
 	 * Pretty print a JSON formatted string. Copied/pasted from http://stackoverflow.com/questions/6054033/pretty-printing-json-with-php
-	 * @deprecated 1.3.0
+	 * @deprecated 1.3.0 use `json_encode($json, JSON_PRETTY_PRINT);` instead
 	 * @param string $json A JSON formatted object definition
 	 * @return string The nicely formatted JSON definition
 	 */
--- a/core/utils.class.inc.php
+++ b/core/utils.class.inc.php
@@ -445,6 +445,8 @@ class Utils
 	 */
 	public static function JSONPrettyPrint($json)
 	{
+	    Utils::Log(LOG_NOTICE, 'Use of deprecated method '.__METHOD__);
+
 	    $result = '';
 	    $level = 0;
 	    $in_quotes = false;
--- a/core/utils.class.inc.php
+++ b/core/utils.class.inc.php
@@ -439,6 +439,7 @@ class Utils
 
 	/**
 	 * Pretty print a JSON formatted string. Copied/pasted from http://stackoverflow.com/questions/6054033/pretty-printing-json-with-php
+	 * @deprecated 1.3.0
 	 * @param string $json A JSON formatted object definition
 	 * @return string The nicely formatted JSON definition
 	 */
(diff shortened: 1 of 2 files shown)


piRGoif:
Seems like a good idea, thanks ! Seems like the \Utils::JSONPrettyPrint method is only used once in Combodo's collectors, but indeed adding a `@deprecated` is a really good idea as this will allow other consumers to migrate !

core/utils.class.inc.php
```diff
@@ -439,6 +439,7 @@ static public function DoPostRequest($sUrl, $aData, $sOptionnalHeaders = null, &
 
 	/**
 	 * Pretty print a JSON formatted string. Copied/pasted from http://stackoverflow.com/questions/6054033/pretty-printing-json-with-php
+	 * @deprecated 1.3.0
```
piRGoif, line 442: When possible it's nice to explain what to use instead O:)

```suggestion
	 * @deprecated 1.3.0 use `json_encode($json, JSON_PRETTY_PRINT);` instead
```

piRGoif (changes requested):

Hipska:
@piRGoif What about replacing the content of the deprecated method with a php native code?

piRGoif (approved):

Hipska:
And maybe also a log warning message?

piRGoif:
Thanks for the deprecated message change !

About changing \Utils::JSONPrettyPrint, I would advise against it as this may introduce behavior change : both codes (Combodo's and PHP) might not produce the same result...
Having a transition period will allow to fallback to the old method if needed.

I'm setting this PR to be reviewed by the dev team !

Hipska:
Good point, but I would still add a log message if this method is still used. Agree?

piRGoif:
> And maybe also a log warning message?

Yes seems like a good idea.
I have an ungoing proposition about logging deprecated calls (Combodo/iTop#193) but it won't apply here as this would introduce dependency to the iTop version where this will be introduced (probably 3.0.0).
So a trigger_error maybe ? We'll have to talk about this with the dev team !

Hipska:
Oh yeah, I was just thinking about a simple `Utils::Log` message.

piRGoif:
>> a trigger_error maybe ?
> Oh yeah, I was just thinking about a simple `Utils::Log` message.

That is an option too indeed, and less "violent". Well noted !

piRGoif (approved):
It was a good idea to add a notice example. Again, the final solution will be chosed by the dev team
I'm approving the code as is for now, and we'll see what the dev team discussion brings to the table :)

piRGoif:
Ref for `JSON_PRETTY_PRINT` flag availability : https://www.php.net/manual/fr/json.constants.php#constant.json-pretty-print

core/utils.class.inc.php
```diff
@@ -439,11 +439,14 @@ static public function DoPostRequest($sUrl, $aData, $sOptionnalHeaders = null, &
 
 	/**
 	 * Pretty print a JSON formatted string. Copied/pasted from http://stackoverflow.com/questions/6054033/pretty-printing-json-with-php
+	 * @deprecated 1.3.0 use `json_encode($value, JSON_PRETTY_PRINT);` instead
```
piRGoif, line 442:
```suggestion
	 * @deprecated 1.3.0 use `json_encode($value, JSON_PRETTY_PRINT);` instead (PHP 5.4.0 required)
```

piRGoif (changes requested):
Just reviewing this PR before showing it to the dev team, I think adding the PHP requirement in the phpdoc would be a good idea

piRGoif (approved):
Thanks !

toolkit/dump_tasks.php
```diff
@@ -113,8 +113,7 @@
 					$aCurrentTaskDefinition['database_table_name'] = "";
 				}
 
-				$sDefinition = json_encode($aCurrentTaskDefinition);
-				echo Utils::JSONPrettyPrint($sDefinition)."\n";
+				echo json_encode($aCurrentTaskDefinition, JSON_PRETTY_PRINT);
```
piRGoif, line 116:
```suggestion
				if (defined('JSON_PRETTY_PRINT') {
					echo json_encode($aCurrentTaskDefinition, JSON_PRETTY_PRINT);
				} else {
					echo Utils::JSONPrettyPrint($sDefinition)."\n";
				}
```

piRGoif:
Thanks, merging !

Hipska:
FYI, i don't think there is still the need to check if the pretty print exists as the collector requires 5.6 minimal by default..

piRGoif:
Thanks for this remark
I'm not sure the collector does a good job checking for PHP Version ? 
This would take some time to check, I'm not sure this is worse the effort ?

Hipska:
The problem is more that the json/csv/sql collectors are loaded/included always. And they set their own php version requirement, even when they are not used. The command used is just `phpversion()`, so I don't think that's giving a impact.

The `Orchestrator::AddCollector` could call a method on the class like `Collector::GetRequirementList` maybe?

PR #12, opened 2021-09-09
N°5158 - Improve robustness toolkit/dump_tasks.php with PHP 7.2+ and 8.0+

Hipska:
The main change in this PR is a fix for when dump_tasks results in 0 objects. In that case the field is not an array, but a `null` value.

Next, I also added the tools to accept `console_log_level` parameter.


toolkit/dump_tasks.php
```diff
@@ -28,6 +28,7 @@
 require_once(APPROOT.'core/restclient.class.inc.php');
 
 $sTaskName = Utils::ReadParameter('task_name', '*');
+Utils::$iConsoleLogLevel = Utils::ReadParameter('console_log_level', Utils::GetConfigurationValue('console_log_level', LOG_WARNING));
```
piRGoif, line 31: In current code base, we have : 
`	static public $iConsoleLogLevel = LOG_INFO;`
So to my opinion this is not equivalent ?

piRGoif (changes requested):

toolkit/dump_tasks.php
Hipska: I took this line of code exactly from here:
https://github.com/Combodo/itop-data-collector-base/blob/60ed0c602ebdca939651eea65557a5ca51cee33d/exec.php#L81

I agree it maybe should be like this on all places?
```suggestion
Utils::$iConsoleLogLevel = Utils::ReadParameter('console_log_level', Utils::GetConfigurationValue('console_log_level', Utils::$iConsoleLogLevel));
```

Hipska: + the `params.distrib.xml` file sets this by default to 6 (INFO), so there's no impact at all..

Molkobain:
@piRGoif @Hipska this PR is in "WIP" state, do we still need something to be changed before review?

Hipska:
Not from my side, its not marked as a draft..

piRGoif:
I opened a discussion on that PR and Thomas answered, but I didn't took the time to take a look since then.
Won't be able to do it this week...
I think at least one dev have to deep dive in this before the review ?

Hipska:
Hi @piRGoif, there were some conflicts in the meantime apparently, so I resolved them..

Hipska:
There is not much deep dive into this, I just re-arranged a switch-case with addition of an if-else to avoid PHP warnings on 7.2+ and fatal error on PHP 8.0+: https://3v4l.org/TEJ4n

toolkit/testconnection.php
```diff
@@ -5,6 +5,8 @@
 require_once(APPROOT.'core/utils.class.inc.php');
 require_once(APPROOT.'core/restclient.class.inc.php');
 
+Utils::$iConsoleLogLevel = Utils::ReadParameter('console_log_level', Utils::GetConfigurationValue('console_log_level', LOG_WARNING));
```
piRGoif, line 8: I imagine that this is set because there are Utils::Log calls somewhere in the called methods stack ? 
I began to search but wasn't able to find :/

There's the same line added in dump_task.php
And this is already present in exec.php...
If we're duplicating code present elsewhere I would prefer to factorize it in a helper method ?

piRGoif (changes requested):

toolkit/testconnection.php
Hipska: Sure, fine for me. But is that to be done in this PR?

piRGoif: As code duplication is done by this PR I would say, yes O:)
You could create a Utils method, maybe InitIConsoleLogLevel ?

.gitignore
```diff
@@ -2,6 +2,7 @@
 composer.lock
 vendor/*
 collectors
+conf/params.local.xml
```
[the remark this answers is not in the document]
Hipska: I don't think that's something you would want. Using that argumentation, You could also replace `composer.lock` by `*.lock` for example..

Anyway, if you really want me to do that, I don't even think your line 6 is needed, since that one is already checked in..

toolkit/testconnection.php
Hipska: Okay, and that one would be called by `Utils::LoadConfig` or be placed instead of this current line we are now discussing?

piRGoif: I was thinking of a simple helper to replace this single line. But we'll see with the rest of the dev team

Hipska: Okay, that sounds more logical indeed.

Hipska: Done.

.gitignore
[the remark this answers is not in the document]
piRGoif: Ok, you've done it in ee3b7cb
I'm closing this conversation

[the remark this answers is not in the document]
piRGoif: Woops sorry my mistake, I picked the wrong conversation ! This one is about the gitignore file sorry
Reopening it !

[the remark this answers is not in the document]
Hipska: No I didn't

piRGoif:
Sorry, we missed technical review today (too much PR to review). Will be reviewed next month !

.gitignore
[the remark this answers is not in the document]
Molkobain: Reviewed today, please just apply the changed proposed by @piRGoif upper in this thread and it will be accepted (technically).
Functional review is next week.

piRGoif:
Technical review was done on april 5th
Modification asked is now done. 
Ready for functional review !

core/utils.class.inc.php
```diff
@@ -108,6 +108,20 @@ static public function CheckParameters($aOptionalParams)
 		return $aUnknownParams;
 	}
 
+	/**
+	 * Init the console log level.
+	 *
+	 * Defaults to LOG_INFO if `console_log_leve` is not configured
```
Molkobain, line 114:
```suggestion
	 * Defaults to LOG_INFO if `console_log_level` is not configured
```

Molkobain (changes requested):
Small typo then ready to merge

Molkobain:
Accepted in functional review

PR #11, opened 2021-03-16
N°5709 - Fix warning when aSkippedAttributes is not correctly populated

Hipska:
The array was only populated on initial run (when data synchro needs to be created) or when synchro needs to be updated.

Better way is to use the method intended for this use.

Note that jsoncollector and SQLCollector are also using this array, it might be that those need an update as well..

PR #16, opened 2022-10-06
N°5600 - Create enhanced collector::CheckColumns method and factorize JSON, CSV and SQL equivalent methods to it

xtophe38:
Purpose of this PR is to standardize and factorize the CheckSynchroColumns / CheckJSONFiedls / CheckSQLColumn methods that exists in csv, json and sql collectors and to slightly alter their behaviour to better handle "optional" attributes.

On the factorization front:
- a new CheckColumns method has been created in the Collector class,
- it replaces the CheckSynchroColumns method of the CSVCollector,
- it replaces the CheckJSONFields method of the JsonCollector as well as a few checks and logs made before its call,
- it replaces the CheckSQLColumn method of the SQLCollector as well as a few checks and logs made before its call.
The 3 legacy methods were doing the same things.

On the "optional" attribute front:
The CheckColumns method skips the checks done on an attribute if this one is optional. This feature significantly reduces the complexity of a collector (like the Azure one) by allowing the json files that describe the synchro policies for classes to hold information about attributes that maybe mandatory in a datamodel and not used in another datamodel.
Consequence is that the json file may hold the description of an attribute, attribute_1, with the option "update" set to 1 even if this attribute_1 is not used in some datamodel, knowing that this attribute is mandatory an another datamodel.

### DIFF
--- a/core/csvcollector.class.inc.php
+++ b/core/csvcollector.class.inc.php
@@ -301,7 +301,6 @@ abstract class CSVCollector extends Collector
 	protected function CheckColumns($aSynchroColumns, $aColumnsToIgnore, $sSource)
 	{
 		Utils::Log(LOG_DEBUG, "[".get_class($this)."] Columns [".var_export($this->aSynchroColumns, true)."]");
-		$aColumnsToIgnore = [];
 		foreach ($this->aFields as $sSynchroColumn => $aDefs) {
 			if (array_key_exists($sSynchroColumn, $this->aSynchroFieldsToDefaultValues) || in_array($sSynchroColumn, $this->aIgnoredSynchroFields)) {
 				$aColumnsToIgnore[] = $sSynchroColumn;
--- a/core/csvcollector.class.inc.php
+++ b/core/csvcollector.class.inc.php
@@ -300,14 +300,14 @@ abstract class CSVCollector extends Collector
 	 */
 	protected function CheckColumns($aSynchroColumns, $aColumnsToIgnore, $sSource)
 	{
-		Utils::Log(LOG_DEBUG, "[".get_class($this)."] Columns [".var_export($this->aSynchroColumns, true)."]");
+		Utils::Log(LOG_DEBUG, "[".get_class($this)."] Columns [".var_export($aSynchroColumns, true)."]");
 		foreach ($this->aFields as $sSynchroColumn => $aDefs) {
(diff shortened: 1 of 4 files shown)


Hipska:
You might want to have a look at #11 as well, it seems related..

piRGoif:
Dicussed with Christophe : will be reviewed next week

Molkobain:
Small refactoring to do by Christophe.
Otherwise, accepted during technical review. Moving this to functional review so we don't wait another month.

core/csvcollector.class.inc.php
```diff
@@ -226,7 +226,15 @@ public function Fetch()
...
-			$this->CheckSynchroColumns();
+			Utils::Log(LOG_DEBUG, "[".get_class($this)."] Columns [".var_export($this->aSynchroColumns, true)."]");
+			$aColumnsToIgnore = [];
+			foreach ($this->aFields as $sSynchroColumn => $aDefs) {
+				if (array_key_exists($sSynchroColumn, $this->aSynchroFieldsToDefaultValues)
+					|| in_array($sSynchroColumn, $this->aIgnoredSynchroFields)) {
+					$aColumnsToIgnore[] = $sSynchroColumn;
+				}
+			}
+			$this->CheckColumns($this->aSynchroColumns, $aColumnsToIgnore, 'csv file');
```
Molkobain, line 237: Method CheckColumn will be overloaded in this class to include lines 229:236

xtophe38: Done

core/csvcollector.class.inc.php
```diff
@@ -306,6 +295,20 @@ protected function Configure($aCsvHeaderColumns)
...
+	protected function CheckColumns($aSynchroColumns, $aColumnsToIgnore, $sSource)
+	{
+		Utils::Log(LOG_DEBUG, "[".get_class($this)."] Columns [".var_export($this->aSynchroColumns, true)."]");
+		foreach ($this->aFields as $sSynchroColumn => $aDefs) {
+			if (array_key_exists($sSynchroColumn, $this->aSynchroFieldsToDefaultValues) || in_array($sSynchroColumn, $this->aIgnoredSynchroFields)) {
+				$aColumnsToIgnore[] = $sSynchroColumn;
+			}
+		}
+
+		parent::CheckColumns($this->aSynchroColumns, $aColumnsToIgnore, 'csv file');
```
Molkobain, line 310:
```suggestion
		parent::CheckColumns($this->aSynchroColumns, $aColumnsToIgnore, $sSource);
```

Shouldn't this rather use the given parameter?

Molkobain (changes requested):

PR #21, opened 2023-01-13
 N°5884 - Fix missing attributes in aSkippedAttribute

accognet:
For some classes, array aSkippedAttribute doesn't contain all the fields -> csv created for import contains fields that don't exist in iTop and the import fails.

### DIFF
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -870,6 +870,7 @@ abstract class Collector
 							if ($this->AttributeIsOptional($sAttCode)) {
 								// Ignore missing optional attributes
 								Utils::Log(LOG_DEBUG, "Comparison: ignoring the missing, but optional, attribute: '$sAttCode'.");
+								$this->aSkippedAttributes[] = $sAttCode;
 								continue;
 							} else {
 								// Missing non-optional attribute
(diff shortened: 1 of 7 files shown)


xtophe38:
Do you have an easy way to add a PHP Unit test ?

xtophe38 (commented):
Reviewed. See comment on Unit Test.

odain-cbd:
Same as Christophe. can you add some tests that cover this case please? does this concern any collector (csv, json, oql)? if so please add a test on each of them?

odain-cbd:
this PR description is hard to understand. this is a technical explanation. not a fonctionnal one.

can you explain when this is happens (which step among data synchro init/collect/synchro itself)? what is the effect? what do you suggest to avoid such an issue?

odain-cbd (commented):
code changes are ok to me. a better explanation would be nice to ease the PR review for others...

odain-cbd (changes requested):
code changes are ok to me. a better explanation would be nice to ease the PR review for others...

xtophe38:
Everything is fine for me.

piRGoif:
Thanks for the review ! We can skip the technical review. Setting to be in the next functional review.

Molkobain (approved):

Molkobain:
Accepted during collector PRs review, you can merge.
Please mind to make a "squash & merge" while keeping the PR number (#xx)

PR #23, opened 2023-02-02
N°5749 - Lack of feedback in case of synchro

odain-cbd:
Sometimes your collector runs successively but part of your data were not updated for below reason:
`  0: Wrong format for DATETIME column 11: 'rrr' does not match the expected format: 'Y-m-d H:i:s' (column skipped)`

This feedback is not visible in collector output nor on iTop side (logs/synchro/replica). which is really painfull and time consuming...

Current PR aims at calling iTop synchro with details in case of log level set to DEBUG.

Curl options are also enhanced to be able to send options that have bool or string values :
			CURLOPT_SSLVERSION => CURL_SSLVERSION_SSLv3,
			CURLOPT_COOKIE => 'itop',
			CURLOPT_COOKIESESSION => true


Hipska:
Why not use debug log flag for that?

odain-cbd:
> Why not use debug log flag for that?

do you mean "--console_log_level" parameter? if so at which level should we switch to details synchro output?

Hipska:
Yes, level DEBUG or INFO I would say.

accognet:
"Synchronize" function in collector class is override in some collectors (odoo, customer invoice and maybe others). It's not a good idea to change its parameters. I prefer logs as proposed by Hipska

core/collector.class.inc.php
```diff
@@ -636,7 +636,15 @@ protected function RemoveDataFiles()
...
-	public function Synchronize($iMaxChunkSize = 0)
+
+	/**
+	 * @param bool $bSynchroDetails
+	 * @param $iMaxChunkSize
+	 *
+	 * @return bool
+	 * @throws \Exception
+	 */
+	public function Synchronize(bool $bSynchroDetails, $iMaxChunkSize = 0)
```
Molkobain, line 647: Discussed during collector PRs review.

As Anne-Catherine pointed out, this method is overloaded by other collectors, so changing its prototype by inserting an argument first will break derived classes.

Also we would like to keep the consistency of debugging by keeping this kind of output in the logs. So we would tend to using `Utils::Log` with a DEBUG level.

Molkobain (changes requested):

odain-cbd:
changes have been made and are covered with tests. that is why I extracted some static logic to call iTop in a service. that I mock in the tests...

core/collector.class.inc.php
```diff
@@ -14,6 +14,8 @@
 //   You should have received a copy of the GNU Affero General Public License
 //   along with this application. If not, see <http://www.gnu.org/licenses/>
 
+require_once(APPROOT.'core/callitopservice.class.inc.php');
+
```
Hipska, line 18: Is this the correct place to add this?

core/collector.class.inc.php
```diff
@@ -69,6 +74,15 @@ public function __construct() {
 		$this->aSkippedAttributes = [];
 	}
 
+	public static function SetCallItopService($oCurrentCallItopService=null){
+		if (null === $oCurrentCallItopService){
+			$oCurrentCallItopService = new CallItopService();
+		}
+
+		Collector::$oCallItopService = $oCurrentCallItopService;
```
Hipska, line 82:
```suggestion
		static::$oCallItopService = $oCurrentCallItopService;
```

core/collector.class.inc.php
```diff
@@ -717,32 +738,10 @@ public function Synchronize($iMaxChunkSize = 0)
...
-
-		$aRawCurlOptions = Utils::GetConfigurationValue('curl_options', array(CURLOPT_SSLVERSION => CURL_SSLVERSION_SSLv3));
-		$aCurlOptions = array();
-		foreach ($aRawCurlOptions as $key => $value) {
-			// Convert strings like 'CURLOPT_SSLVERSION' to the value of the corresponding define i.e CURLOPT_SSLVERSION = 32 !
-			$iKey = (!is_numeric($key)) ? constant((string)$key) : (int)$key;
-			$iValue = (!is_numeric($value)) ? constant((string)$value) : (int)$value;
-			$aCurlOptions[$iKey] = $iValue;
+		if (null === Collector::$oCallItopService){
+			self::SetCallItopService();
```
Hipska, line 742:
```suggestion
		if (null === static::$oCallItopService){
			static::SetCallItopService();
```

core/collector.class.inc.php
```diff
@@ -717,32 +738,10 @@ public function Synchronize($iMaxChunkSize = 0)
...
 		}
-		$aCurlOptions[CURLOPT_CONNECTTIMEOUT] = $iCurrentTimeOut;
-		$aCurlOptions[CURLOPT_TIMEOUT] = $iCurrentTimeOut;
-
-		return Utils::DoPostRequest($sUrl, $aData, '', $aResponseHeaders, $aCurlOptions);
+		return Collector::$oCallItopService->CallItopViaHttp($sUri, $aAdditionalData, $iTimeOut);
```
Hipska, line 744:
```suggestion
		return static::$oCallItopService->CallItopViaHttp($sUri, $aAdditionalData, $iTimeOut);
```

core/callitopservice.class.inc.php
```diff
@@ -0,0 +1,34 @@
...
+		);
+
+		$aRawCurlOptions = Utils::GetConfigurationValue('curl_options', array(CURLOPT_SSLVERSION => CURL_SSLVERSION_SSLv3));
+		$aCurlOptions = array();
+		foreach ($aRawCurlOptions as $key => $value) {
+			// Convert strings like 'CURLOPT_SSLVERSION' to the value of the corresponding define i.e CURLOPT_SSLVERSION = 32 !
+			$iKey = (!is_numeric($key)) ? constant((string)$key) : (int)$key;
+			$iValue = (!is_numeric($value)) ? constant((string)$value) : (int)$value;
+			$aCurlOptions[$iKey] = $iValue;
+		}
```
Hipska, line 28: This is wrong, not all cURL options are integers, see https://www.php.net/manual/en/function.curl-setopt.php

Hipska (changes requested):

core/callitopservice.class.inc.php
Molkobain: It's wrong indeed, but it's not a change from this PR, Olivier just extracted it in a dedicated method.

core/callitopservice.class.inc.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+class CallItopService
```
Molkobain, line 3: Please add a `@since 1.3.0` so people extending the collector know since when this is available if they want to use the service directly.

core/collector.class.inc.php
```diff
@@ -69,6 +74,15 @@ public function __construct() {
 		$this->aSkippedAttributes = [];
 	}
 
+	public static function SetCallItopService($oCurrentCallItopService=null){
```
Molkobain, line 77: I can't add suggestion but please:
- Add `@since`
- Type the $oCurrentCallItopService parameter
- (Surround "=" with spaces)

core/collector.class.inc.php
Molkobain: +1 this allows other collectors to potentially overload the way the remote iTop is called.

core/collector.class.inc.php
Molkobain: +1

core/collector.class.inc.php
Molkobain: +1 see other comment

Molkobain (changes requested):

core/collector.class.inc.php
odain-cbd: why not? it seems simpler to me to include it only once here. instead of all the places where collector.php is required also (exec.php + all the tests).

core/collector.class.inc.php
odain-cbd: +1

core/collector.class.inc.php
odain-cbd: +1

core/callitopservice.class.inc.php
odain-cbd: i will not include your proposal in current PR, Hispka. please propose it apart in another one. still it seems a good one.

core/collector.class.inc.php
Hipska: Yes, that is also done for `ioexception.class.inc.php` and similar.

core/callitopservice.class.inc.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+/**
+ * @since 1.3.0
+ */
+class CallItopService
+{
+	public function CallItopViaHttp($sUri, $aAdditionalData, $iTimeOut = -1)
+	{
+		$iCurrentTimeOut = ($iTimeOut !== -1) ? $iTimeOut : (int)Utils::GetConfigurationValue('itop_synchro_timeout', 600); // timeout in seconds, for a synchro to run
```
Hipska, line 10: Would it make sense to reverse the arguments?
```suggestion
		$iCurrentTimeOut = ($iTimeOut === -1) ? (int)Utils::GetConfigurationValue('itop_synchro_timeout', 600) : $iTimeOut; // timeout in seconds, for a synchro to run
```

core/callitopservice.class.inc.php
```diff
@@ -0,0 +1,37 @@
...
+		foreach ($aRawCurlOptions as $key => $value) {
+			// Convert strings like 'CURLOPT_SSLVERSION' to the value of the corresponding define i.e CURLOPT_SSLVERSION = 32 !
+			$iKey = (!is_numeric($key)) ? constant((string)$key) : (int)$key;
+			$iValue = (!is_numeric($value)) ? constant((string)$value) : (int)$value;
+			$aCurlOptions[$iKey] = $iValue;
+		}
+		$aCurlOptions[CURLOPT_CONNECTTIMEOUT] = $iCurrentTimeOut;
+		$aCurlOptions[CURLOPT_TIMEOUT] = $iCurrentTimeOut;
+
+		return Utils::DoPostRequest($sUrl, $aData, '', $aResponseHeaders, $aCurlOptions);
```
Hipska, line 35: `$aResponseHeaders` doesn't seem to be actually used..
```suggestion
		return Utils::DoPostRequest($sUrl, $aData, '', null, $aCurlOptions);
```

core/collector.class.inc.php
Hipska: Not resolved..

core/collector.class.inc.php
```diff
@@ -717,32 +737,10 @@ public function Synchronize($iMaxChunkSize = 0)
...
-		$aRawCurlOptions = Utils::GetConfigurationValue('curl_options', array(CURLOPT_SSLVERSION => CURL_SSLVERSION_SSLv3));
-		$aCurlOptions = array();
-		foreach ($aRawCurlOptions as $key => $value) {
-			// Convert strings like 'CURLOPT_SSLVERSION' to the value of the corresponding define i.e CURLOPT_SSLVERSION = 32 !
-			$iKey = (!is_numeric($key)) ? constant((string)$key) : (int)$key;
-			$iValue = (!is_numeric($value)) ? constant((string)$value) : (int)$value;
-			$aCurlOptions[$iKey] = $iValue;
+		if (null === static::$oCallItopService){
+			self::SetCallItopService(new CallItopService());
 		}
```
Hipska, line 742: No added value to call the method
```suggestion
		if (null === static::$oCallItopService){
			static::$oCallItopService = new CallItopService();
		}
```
Otherwise
```suggestion
		if (null === static::$oCallItopService){
			static::SetCallItopService(new CallItopService());
		}
```

Hipska (changes requested):

core/collector.class.inc.php
Molkobain: This suggestion is missing from the applied ones.

core/collector.class.inc.php
```diff
@@ -717,32 +737,10 @@ public function Synchronize($iMaxChunkSize = 0)
...
-
-		$aRawCurlOptions = Utils::GetConfigurationValue('curl_options', array(CURLOPT_SSLVERSION => CURL_SSLVERSION_SSLv3));
-		$aCurlOptions = array();
-		foreach ($aRawCurlOptions as $key => $value) {
-			// Convert strings like 'CURLOPT_SSLVERSION' to the value of the corresponding define i.e CURLOPT_SSLVERSION = 32 !
-			$iKey = (!is_numeric($key)) ? constant((string)$key) : (int)$key;
-			$iValue = (!is_numeric($value)) ? constant((string)$value) : (int)$value;
-			$aCurlOptions[$iKey] = $iValue;
+		if (null === static::$oCallItopService){
+			self::SetCallItopService(new CallItopService());
```
Molkobain, line 741:
```suggestion
			static::SetCallItopService(new CallItopService());
```

Molkobain (changes requested):
2 missing changes, then ok to me.

core/callitopservice.class.inc.php
odain-cbd: +1

core/collector.class.inc.php
odain-cbd: +1

core/callitopservice.class.inc.php
odain-cbd: 3rd parameter is passed by reference. we cannot pass null value directly. but i moved unused variable closer to DoPostRequest call.

Molkobain (approved):

Hipska (changes requested):

core/callitopservice.class.inc.php
odain-cbd: your proposal is nice and quick to test. moreover it is the good momentum as there is an ongoing collector release coming.... I included it.

Molkobain (approved):

test/CallItopServiceTest.php
```diff
@@ -0,0 +1,42 @@
...
+class CallItopServiceTest extends TestCase
+{
+	/** @var \CallItopService $oCallItopService */
+	private $oCallItopService;
+
+	public function testGetCurlOptions(){
+		$aRawCurlOptions = [
+			CURLOPT_SSLVERSION => CURL_SSLVERSION_SSLv3,
+			CURLOPT_COOKIE => 'itop',
+			CURLOPT_COOKIESESSION => true
```
Hipska, line 21: Also add an option with the name as string and also the value as string like `'CURL_SSLVERSION_SSLv3'`.

core/callitopservice.class.inc.php
```diff
@@ -0,0 +1,44 @@
...
+				'auth_user' => Utils::GetConfigurationValue('itop_login', ''),
+				'auth_pwd' => Utils::GetConfigurationValue('itop_password', ''),
+			),
+			$aAdditionalData
+		);
+
+		// timeout in seconds, for a synchro to run
+		$iCurrentTimeOut = ($iTimeOut === -1) ? (int)Utils::GetConfigurationValue('itop_synchro_timeout', 600) : $iTimeOut;
+		$aRawCurlOptions = Utils::GetConfigurationValue('curl_options', [CURLOPT_SSLVERSION => CURL_SSLVERSION_SSLv3]);
+		$aCurlOptions = $this->GetCurlOptions($aRawCurlOptions, $iCurrentTimeOut);
```
Hipska, line 25: This might be better located in Utils? The naming is also not ideal.
```suggestion
		$aCurlOptions = Utils::ProcessCurlOptions($aRawCurlOptions, $iCurrentTimeOut);
```

core/collector.class.inc.php
Hipska: Still no response..

Hipska (changes requested):

core/callitopservice.class.inc.php
odain-cbd: +1 i did it and reused added method

Hipska:
Please don't ask for a new review if you didn't reply to my previous review remarks/questions..

odain-cbd:
> Please don't ask for a new review if you didn't reply to my previous review remarks/questions..

sorry if i missed sth. please point it out next time. after few rounds of reviews github PR UI becomes messy to me.

Hipska:
Yes, see here: https://github.com/Combodo/itop-data-collector-base/pull/23#discussion_r1108311621

Mainly because you marked it already as resolved..

test/CallItopServiceTest.php
Hipska: The value also needs a string constant test.

core/collector.class.inc.php
odain-cbd: IOException class should be found everywhere this kind of exception is thrown and catched. 

Regarding CallItopService this class is used  only within dedicated test and collector.php file. that is why I did not spread this require everywhere. it seems cleaner to me. 
collector.php is spread already. it does the job for callitopservice.class.inc.php as well. no?

test/CallItopServiceTest.php
odain-cbd: correction made. as I moved that code in Utils class  test class/file have been renamed as well

odain-cbd: Test has been rewritten to highlight all usecases covered. for me the one you mentioned is tested. cf line 25 ('string key/ string value') inside UtilsTest.php.

Hipska: No, the code also supports the value to be like `'CURL_SSLVERSION_SSLv3'` and the expected result should be `3`.

odain-cbd: ok i got it. sorry.

core/collector.class.inc.php
Hipska: It's fine for me if others in the team agree with this added inconsistency.

Molkobain: It is fine by me as well to leave this require here as it is the only place it is used. That's why I didn't comment before.

Molkobain (approved):

PR #27, opened 2023-02-15
N°5995 - Fix PHP warning on lookup when table is empty

accognet:
Avoid the message :
PHP Warning: Invalid argument supplied for foreach() in .... core/lookuptable.class.inc.php on line 59
when the requested table is empty

### DIFF
--- a/core/lookuptable.class.inc.php
+++ b/core/lookuptable.class.inc.php
@@ -56,45 +56,65 @@ class LookupTable
 		}
 		$aRes = $oRestClient->Get($sClass, $sOQL, implode(',', $aRestFields));
 		if ($aRes['code'] == 0) {
-			foreach ($aRes['objects'] as $sObjKey => $aObj) {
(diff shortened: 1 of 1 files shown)


Hipska:
Could also just have done `foreach ((array)$aRes['objects'] as $sObjKey => $aObj)`

Molkobain:
Indeed the fix proposed by Thomas would be nicer as it avoids the extra indentation. I'm in favor for it as well.

Also, your formatter changed the brackets position to something not in the convention. Please revert these and check the changes in your IDE before committing.

Hipska (approved):

xtophe38 (approved):
Tested on Azure collector. Works fine.

Molkobain:
Fine by me as well, you can
  * Squash & merge the PR
  * Resolve the corresponding ticket
  * Update the changelogs of the base, LDAP and vSphere collectors

PR #28, opened 2023-02-15
N°5996 - During creation of csv file, in function lookup remove warning if initial field is empty

accognet:
During creation of csv file, in function lookup remove warning if initial field is empty.

Example :
During the collection process, in my initial data, oslicence_id is empty. I have a lookup on this field. This lookup does not find any data and it's normal. But in the logs, there is a warning :
[Warning]       No mapping found with key: '47_', 'oslicence_id' will be set to zero.

I want to remove this warning

### DIFF
--- a/core/utils.class.inc.php
+++ b/core/utils.class.inc.php
@@ -29,6 +29,7 @@ class Utils
 	static public $oCollector = "";
 	static protected $oConfig = null;
 	static protected $aConfigFiles = array();
+	static protected $iMockLogLevel = LOG_ERR;
 
 	static protected $oMockedLogger;
 
@@ -136,7 +137,7 @@ class Utils
 	{
 		//testing only LOG_ERR
 		if (self::$oMockedLogger) {
-			if ($iPriority <= self::$iConsoleLogLevel && $iPriority <= LOG_ERR) {
+			if ($iPriority <= self::$iConsoleLogLevel && $iPriority <= self::$iMockLogLevel) {
 				var_dump($sMessage);
 				self::$oMockedLogger->Log($iPriority, $sMessage);
 			}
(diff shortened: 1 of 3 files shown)


Hipska:
This change is not needed, you only need to set `$this->bIgnoreMappingErrors` to `true`.

accognet:
You are right, if you set $this->bIgnoreMappingErrors to true, the warning goes away, but there is still a debug message. And for me it's not the same if initial field is empty or not.

Hipska:
Ok, I understand. But then I would add the logic more to line 140 where the current message is emitted.

accognet:
It's an option, but at the beginning of the function, this avoids looking for something that does not exist.

odain-cbd:
it would be nice to add to cover lookup function with tests. nominal ones and one regarding the warning you are currently removing. that way our CI will also test this part of the code on different versions of PHP...

core/utils.class.inc.php
```diff
@@ -213,9 +214,10 @@ private static function CreateEventIssue($sMessage)
 		$oClient->Create("EventIssue", $aFields, 'create event issue from collector $sCollectorName execution.');
 	}
 
-	static public function MockLog($oMockedLogger)
```
odain-cbd, line 216: test are broken. below signature would fix them i guess:
static public function MockLog($oMockedLogger, $iMockLogLevel=LOG_ERR)

accognet: Done

Molkobain (approved):

odain-cbd (approved):
just check it is fine for our ci as well before merging :)

great that you added so much tests. our coverage was really bad on this part of the code

PR #29, opened 2023-02-22
N°6012 - Use authentication via tokens to reach iTop

odain-cbd:
collector-base communicate with iTop through login/password (rest/synchro). this PR aims at passing tokens instead to authenticate. for that purpose iTop should have authent-token extension.

here are the XML configurations covered:

* authentication via login/password

<?xml version="1.0" encoding="UTF-8"?>

<parameters>
  <itop_url>https://localhost/iTop</itop_url>
  <itop_login>admin1</itop_login>
  <itop_password>admin2</itop_password>

POST parameters will include 
['auth_user'=> 'admin1', 'auth_pwd2'=>'admin']

in URL login_form remains at value 'form'.

 * authentication via token (next version of authent-token)

<?xml version="1.0" encoding="UTF-8"?>

<parameters>
  <itop_url>https://localhost/iTop</itop_url>
  <itop_login>admin1</itop_login>
  <itop_password>admin2</itop_password>
  <itop_token>admin_token</token>

POST parameters will include  ['token'=> 'admin_token']
in URL login_form is changed to value 'token'.

 * authentication via token (first version of authent-token)

<?xml version="1.0" encoding="UTF-8"?>

<parameters>
  <itop_url>https://localhost/iTop</itop_url>
  <itop_login>admin1</itop_login>
  <itop_password>admin2</itop_password>
  <itop_rest_token>admin_token</rest-token>

POST parameters will include  ['rest-token'=> 'admin_token']
in URL login_form is changed to value 'rest-token'.

To override login-form in URL you still can do it via XML itop_login_form parameter.

### DIFF
--- a/conf/params.distrib.xml
+++ b/conf/params.distrib.xml
@@ -4,7 +4,9 @@
   <itop_url>https://localhost/iTop</itop_url>
   <itop_login>admin</itop_login>
   <itop_password>admin</itop_password>
-  
+  <rest-token/>
+  <token/>
+
   <!-- console_log_level: level of logging to console (std output)
   -1 : none, nothing will be logged to the console
    0 : System wide emergency errors only (LOG_EMERG)
@@ -68,11 +70,11 @@
   <NAME_OF_THE_OPTION1>VALUE_OF_THE_OPTION1</NAME_OF_THE_OPTION1>
   <NAME_OF_THE_OPTION2>VALUE_OF_THE_OPTION2</NAME_OF_THE_OPTION2>
   etc...
-  
+
   Where NAME_OF_THE_OPTIONx and VALUE_OF_THE_OPTIONx are either:
   - The numeric value of the option
   - or the string representation of the corresponding PHP "define" (case sensitive)
-  
+
   The four examples below are equivalent:
    <CURLOPT_SSLVERSION>CURL_SSLVERSION_TLSv1_2</CURLOPT_SSLVERSION>
   or
@@ -81,7 +83,7 @@
    <32>CURL_SSLVERSION_TLSv1_2</32>
   or
    <32>6</32>
-   
+
   Since in PHP we have:
   define ('CURLOPT_SSLVERSION', 32);
   define ('CURL_SSLVERSION_TLSv1_2', 6);
--- a/conf/params.distrib.xml
+++ b/conf/params.distrib.xml
@@ -6,6 +6,7 @@
   <itop_password>admin</itop_password>
   <itop_rest_token/>
   <itop_token/>
+  <itop_login_mode/>
 
   <!-- console_log_level: level of logging to console (std output)
   -1 : none, nothing will be logged to the console
--- a/conf/params.distrib.xml
+++ b/conf/params.distrib.xml
@@ -4,7 +4,6 @@
   <itop_url>https://localhost/iTop</itop_url>
   <itop_login>admin</itop_login>
   <itop_password>admin</itop_password>
-  <itop_rest_token/>
   <itop_token/>
   <itop_login_mode/>
 
--- a/conf/params.distrib.xml
+++ b/conf/params.distrib.xml
@@ -4,8 +4,8 @@
   <itop_url>https://localhost/iTop</itop_url>
   <itop_login>admin</itop_login>
   <itop_password>admin</itop_password>
-  <rest-token/>
-  <token/>
+  <itop_rest_token/>
+  <itop_token/>
 
   <!-- console_log_level: level of logging to console (std output)
   -1 : none, nothing will be logged to the console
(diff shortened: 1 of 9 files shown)


conf/params.distrib.xml
```diff
@@ -4,7 +4,9 @@
   <itop_url>https://localhost/iTop</itop_url>
   <itop_login>admin</itop_login>
   <itop_password>admin</itop_password>
-  
+  <rest-token/>
+  <token/>
```
Hipska, line 8: Should maybe better be renamed to `itop_token` to be streamlined with other options and some data-collectors might already have `token` in use.

odain-cbd: good point i will change

core/dopostrequestservice.class.inc.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+/**
+ * @since 1.3.0 N°6012
+ */
+class DoPostRequestService
+{
+	public function DoPostRequest($sUrl, $aData, $sOptionnalHeaders = null, &$aResponseHeaders = null, $aCurlOptions = []){
```
Molkobain, line 8: I don't get the point of this method, doesn't seem overloaded.

conf/params.distrib.xml
Molkobain: Seems like this part has not been reworked.

**Point 1**
That being said having both "token" and "rest-token" is really confusing for people not knowing about the first beta version of the `authent-token` module.

Do you have a real use case where the legacy form would already be deployed?

Otherwise, as this (authent-token) has not really be published yet and is not spread out, I would rather not support the old "token" at all as it bring more confusion and technical debt than benefits.
The few people (only 1 customer is **maybe** using it in a CEX integration; and maybe some people using the `combodo-intercom-integration`) will be forced to upgrade this module but it is a totally understandable tradeoff IMO. Just needs to be documented along with all the new changes of the data collector.

**Point 2**
I have to dig into the new version of `authent-token`, but is `rest-token` the name of the param in which lies the token? Why not use something like `authent_[rest_]token` to be consistent with the other parameters. It would also make it explicit that is it for authentication.

core/utils.class.inc.php
```diff
@@ -278,6 +294,52 @@ static public function GetConfigurationValue($sCode, $defaultValue = '')
 		return $value;
 	}
 
+	/**
+	 * @since 1.3.0 N°6012
+	 */
+	static public function GetCredentials() : array {
+		$sToken = Utils::GetConfigurationValue('itop_token', '');
+		if (! empty($sToken)){
```
Molkobain, line 302: Don't use `empty()` to detect if a string is empty or not, as it returns true for `'0'`, use `strlen() > 0` instead (not in the documented coding standards I agree, but discussed internally a while ago. I just added it [here](https://www.itophub.io/wiki/page?id=latest:customization:coding_standards#test_if_a_string_is_empty_with_empty))
_Mind the other usages below_

Molkobain (changes requested):

core/dopostrequestservice.class.inc.php
odain-cbd: this class is used to be able to mock Utils::DoPostRequest and test further parts of the collector that use this static method.

odain-cbd: cf :

static public function DoPostRequest($sUrl, $aData, $sOptionnalHeaders = null, &$aResponseHeaders = null, $aCurlOptions = [])
	{
		if (self::$oMockedDoPostRequestService) {
			return self::$oMockedDoPostRequestService->DoPostRequest($sUrl, $aData, $sOptionnalHeaders, $aResponseHeaders, $aCurlOptions);
		}

core/utils.class.inc.php
odain-cbd: ok done

conf/params.distrib.xml
odain-cbd: in next version of authent-token we decided to rename legacy login_mode 'rest-token' to 'token'. for backward compatibility we supported both. let's discuss this apart...

core/utils.class.inc.php
Hipska: It doesn't care much if a token never will be just `'0'`.

conf/params.distrib.xml
```diff
@@ -4,7 +4,10 @@
   <itop_url>https://localhost/iTop</itop_url>
   <itop_login>admin</itop_login>
   <itop_password>admin</itop_password>
-  
+  <itop_rest_token/>
```
Hipska, line 7: Maybe also remove `itop_rest_token` from sample config as it is not to be actively used?

odain-cbd: ok done!

PR #30, opened 2023-03-06
N°6420 dump_tasks : more info

piRGoif:
Change the table header so it is easier to spot iTop URL
Add datasource id as prefix, and remove useless description.

Old log : (87 chars max width)

```
λ php dump_tasks.php
Command line utility to generate the JSON representation of an iTop SynchroDataSource.

Usage: php dump_tasks.php --task_name="selected task name"

There are 2 SynchroDataSource defined on the iTop server (http://localhost/itop-dev):
+--------------------------------+----------------------------------------------------+
|            Name                |                    Description                     |
+--------------------------------+----------------------------------------------------+
| Persons iTop to iTop           |                                                    |
| Test                           | aaaa                                               |
+--------------------------------+----------------------------------------------------+
iTop REST/API version: 1.2
```

New log : (96 char max width)

```
λ php dump_tasks.php                                                                             
Command line utility to generate the JSON representation of an iTop SynchroDataSource.           
                                                                                                 
Usage: php dump_tasks.php --task_name="selected task name"                                       
                                                                                                 
iTop: http://localhost/itop-dev                                                                  
   2 SynchroDataSource objects defined                                                           
+-----+--------------------------------------------------------------------------------------+   
|  Id |             Name                                                                     |   
+-----+--------------------------------------------------------------------------------------+   
|   2 | Persons iTop to iTop                                                                 |   
|   3 | Test                                                                                 |   
+-----+--------------------------------------------------------------------------------------+   
iTop REST/API version: 1.2                                                                                                                   
```

Note : the URL is colored to be easier to see.
Can be seen on this screenshot : 
image

### DIFF
--- a/toolkit/dump_tasks.php
+++ b/toolkit/dump_tasks.php
@@ -63,7 +63,7 @@ if ($sTaskName == '*') {
 		echo "+--------+--------------------------------+----------------------------------------------------+\n";
 		foreach ($aResult['objects'] as $aValues) {
 			$aCurrentTaskDefinition = $aValues['fields'];
-			echo sprintf("| %-6.6s | %-30.30s | %-50.50s |\n",
+			echo sprintf("| %6d | %-30.30s | %-50.50s |\n",
 				$aValues['key'],
 				$aCurrentTaskDefinition['name'],
 				$aCurrentTaskDefinition['description']
--- a/toolkit/dump_tasks.php
+++ b/toolkit/dump_tasks.php
@@ -69,7 +69,7 @@ if ($sTaskName == '*') {
 				$aCurrentTaskDefinition['description']
 			);
 		}
-		echo "+-------+--------------------------------+----------------------------------------------------+\n";
+		echo "+--------+--------------------------------+----------------------------------------------------+\n";
 	}
 	$sMaxVersion = $oRestClient->GetNewestKnownVersion();
 	echo "iTop REST/API version: $sMaxVersion\n";
--- a/toolkit/dump_tasks.php
+++ b/toolkit/dump_tasks.php
@@ -44,31 +44,32 @@ if ($sTaskName == '*') {
 		exit - 1;
 	}
 
-	echo "iTop $sITopUrl, ";
+	echo "iTop: $sITopUrl\n";
 	if (empty($aResult['objects'])) {
-		echo "no SynchroDataSource defined.\n";
+		echo "   no SynchroDataSource defined\n";
 	} else {
 		switch (count($aResult['objects'])) {
 			case 1:
-				echo "1 SynchroDataSource defined:\n";
+				echo "   1 SynchroDataSource defined\n";
 				break;
 
 			default:
-				echo count($aResult['objects'])." SynchroDataSource defined:\n";
+				echo '   '.count($aResult['objects'])." SynchroDataSource objects defined\n";
 				break;
 		}
 
-		echo "+--------------------------------+----------------------------------------------------+\n";
-		echo "|            Name                |                    Description                     |\n";
-		echo "+--------------------------------+----------------------------------------------------+\n";
+		echo "+--------+--------------------------------+----------------------------------------------------+\n";
+		echo "|   Id   |            Name                |                    Description                     |\n";
+		echo "+--------+--------------------------------+----------------------------------------------------+\n";
 		foreach ($aResult['objects'] as $aValues) {
 			$aCurrentTaskDefinition = $aValues['fields'];
-			echo sprintf("| %-30.30s | %-50.50s |\n",
-				$aValues['key'].':'.$aCurrentTaskDefinition['name'],
+			echo sprintf("| %-6.6s | %-30.30s | %-50.50s |\n",
+				$aValues['key'],
+				$aCurrentTaskDefinition['name'],
 				$aCurrentTaskDefinition['description']
 			);
 		}
-		echo "+--------------------------------+----------------------------------------------------+\n";
+		echo "+-------+--------------------------------+----------------------------------------------------+\n";
 	}
 	$sMaxVersion = $oRestClient->GetNewestKnownVersion();
 	echo "iTop REST/API version: $sMaxVersion\n";
--- a/toolkit/dump_tasks.php
+++ b/toolkit/dump_tasks.php
@@ -44,7 +44,7 @@ if ($sTaskName == '*') {
 		exit - 1;
 	}
 
-	echo "iTop: $sITopUrl\n";
+	echo "iTop: \033[34m{$sITopUrl}\033[0m\n";
 	if (empty($aResult['objects'])) {
 		echo "   no SynchroDataSource defined\n";
 	} else {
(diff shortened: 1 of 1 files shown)


toolkit/dump_tasks.php
```diff
@@ -44,16 +44,17 @@
...
-		switch (count($aResult['objects'])){
+		switch (count($aResult['objects'])) {
 			case 1:
-				echo "There is 1 SynchroDataSource defined on the iTop server ($sITopUrl):\n";
+				echo "1 SynchroDataSource defined:\n";
 				break;
 
 			default:
-				echo "There are ".count($aResult['objects'])." SynchroDataSource defined on the iTop server ($sITopUrl):\n";
+				echo count($aResult['objects'])." SynchroDataSource defined:\n";
```
Hipska, line 57:
```suggestion
				echo count($aResult['objects'])." SynchroDataSource's defined:\n";
```

Otherwise this switch and case statement is not needed anymore..

toolkit/dump_tasks.php
```diff
@@ -62,7 +63,10 @@
 		echo "+--------------------------------+----------------------------------------------------+\n";
 		foreach ($aResult['objects'] as $aValues) {
 			$aCurrentTaskDefinition = $aValues['fields'];
-			echo sprintf("| %-30.30s | %-50.50s |\n", $aCurrentTaskDefinition['name'], $aCurrentTaskDefinition['description']);
+			echo sprintf("| %-30.30s | %-50.50s |\n",
+				$aValues['key'].':'.$aCurrentTaskDefinition['name'],
```
Hipska, line 67: Why is the Name value changed? Does this new notation needs to be used as value for the `--task_name` parameter?

piRGoif: id is added as a suffix, to clear the situation when having multiple datasource with very long names
task_name param value should still be the sole datasource name. Ideally I would add a task_key param to specify the source id instead of name, but maybe in a next pr.

toolkit/dump_tasks.php
piRGoif: Oh you're right ! 
But why the apos ?
As it is unusual to make a plural out of an iTop object class name, I would suggest "... SynchroDataSource objects defined". What do you think ?

toolkit/dump_tasks.php
Hipska: @Molkobain indeed, that’s what I was referring to.

toolkit/dump_tasks.php
Hipska: That’s fine as well…

piRGoif:
I like the idea of a new column, will change it

toolkit/dump_tasks.php
piRGoif: Are you referring to Molkobain's comment below about moving the id to a new column ? 
If yes, do you agree on the solution ?

Hipska: Yes, it doesn't make sense to change the name of the datasource being displayed if it can't be used like that. A separate column is indeed better.

toolkit/dump_tasks.php
piRGoif: Done in 1fed14f1

piRGoif:
Modifications pushed, description updated

Hipska (approved):

toolkit/dump_tasks.php
```diff
@@ -44,27 +44,32 @@
...
-		echo "+--------------------------------+----------------------------------------------------+\n";
-		echo "|            Name                |                    Description                     |\n";
-		echo "+--------------------------------+----------------------------------------------------+\n";
+		echo "+--------+--------------------------------+----------------------------------------------------+\n";
+		echo "|   Id   |            Name                |                    Description                     |\n";
+		echo "+--------+--------------------------------+----------------------------------------------------+\n";
 		foreach ($aResult['objects'] as $aValues) {
 			$aCurrentTaskDefinition = $aValues['fields'];
-			echo sprintf("| %-30.30s | %-50.50s |\n", $aCurrentTaskDefinition['name'], $aCurrentTaskDefinition['description']);
+			echo sprintf("| %-6.6s | %-30.30s | %-50.50s |\n",
```
Hipska, line 66: Right-justify integers might look better..
```suggestion
			echo sprintf("| %6d | %-30.30s | %-50.50s |\n",
```

piRGoif: Thanks, nice suggestion ! 
Description updated accordingly

toolkit/dump_tasks.php
```diff
@@ -44,27 +44,32 @@
...
-		switch (count($aResult['objects'])){
+		switch (count($aResult['objects'])) {
 			case 1:
-				echo "There is 1 SynchroDataSource defined on the iTop server ($sITopUrl):\n";
+				echo "   1 SynchroDataSource defined\n";
 				break;
 
 			default:
-				echo "There are ".count($aResult['objects'])." SynchroDataSource defined on the iTop server ($sITopUrl):\n";
+				echo '   '.count($aResult['objects'])." SynchroDataSource objects defined\n";
```
Molkobain, line 57:
```suggestion
				echo "   ".count($aResult['objects'])." SynchroDataSource objects defined\n";
```

Molkobain (approved):

toolkit/dump_tasks.php
piRGoif: Why ? 
I tend to use `"` instead of `'` as string delimiters only when necessary.
Here in the second part of the string we have `\n` that make `"` delimiters mandatory, but in the first part we only have spaces so no need to use `"` right ?

Molkobain: I tend to favor consistency in this kind of thing, especially as there is no performance difference.

piRGoif: Ok
`"` are kind of dangerous as they are really different than `'`, but here the risk is very not significant.

Hipska: Could also use `PHP_EOL` to reduce the need for `"` ;-)

Hipska:
Aslo, does it really need space for 6 digits? I think 4 could be enough as well. Or are there systems with more than 10000 configured data synchro's?

piRGoif:
> Also, does it really need space for 6 digits? I think 4 could be enough as well.

I think you're right ! We should talk about this in the review.
I'll switch back to 4 char in the meantime.

Hipska (approved):

Molkobain:

piRGoif:
Functional review : remove description so that name won't be truncated anymore. Also 3 chars would be enough for the ID field.
Ideally, we should be able to dump task using whether its name or its id O:)

Hipska:
If you're going to remove description, consider putting target class in its place.

piRGoif:
> If you're going to remove description, consider putting target class in its place.

You mean the SynchroDataSource.scope_class field value ? 
Do you have real examples where it would benefit to have both the datasynchro name and target class ?

Hipska (approved):

Hipska:
Yeah, when the synchro name isn't descriptive, see your example ID:3 "Test".

Molkobain:
Accepted during functional review:
- DataSynchro class won't be added to the table, would be in another PR.

PR #31, opened 2023-03-17
N°6109 Fix datatable name removed in dump_task

piRGoif:
This is causing issue when creating the datasynchro on iTop with a custom datatable name, then dumping it and using the JSON in a collector.

Regression introduced in 595abed

### DIFF
--- a/toolkit/dump_tasks.php
+++ b/toolkit/dump_tasks.php
@@ -92,11 +92,6 @@ if ($sTaskName == '*') {
 				$aCurrentTaskDefinition['user_id'] = '$synchro_user$';
 				$aCurrentTaskDefinition['notify_contact_id'] = '$contact_to_notify$';
 
-
-				if (isset($aCurrentTaskDefinition['database_table_name'])) {
-					$aCurrentTaskDefinition['database_table_name'] = "";
-				}
-
 				echo json_encode($aCurrentTaskDefinition, JSON_PRETTY_PRINT);
 			}
 		} else {


Hipska:
What is the issue one is having because of that? I see a lot of the collectors have this value as an empty string.

If you now use dump_tasks to update your collector code, you will again need to manually remove the table name before comitting? (Maybe only make empty if the current table is following the default naming scheme?)

piRGoif:
The problems happens when creating a new collector. After having creating the datasource in iTop, doing a dump, and launching the collector for the first time.
If the datasource has a custom data table name, then the JSON won't, and then the collector will change the datasource to align the data table name. It won't create the corresponding table :( This is another bug.

Letting the data table name empty in the JSON is a good idea if you want to distribute widely your collector : you will reduce the odds of having a collision with the data table name.
But it's not always necessary...

Hipska:
> If the datasource has a custom data table name, then the JSON won't, and then the collector will change the datasource to align the data table name.

I don't see this behaviour. If the datasource exists and in the JSON there is no table name, nothing will happen IMO.

> Letting the data table name empty in the JSON is a good idea if you want to distribute widely your collector : you will reduce the odds of having a collision with the data table name.
> But it's not always necessary...

Correct, that's why my suggestion is to empty the table name if it is following the [default naming scheme](https://github.com/Combodo/iTop/blob/cd8dd596416d502a150dcfe0f95ab202ef9c0cb2/synchro/synchrodatasource.class.inc.php#L908-L928)...

piRGoif:
> I don't see this behaviour.

I did this PR right after having this bug, so I'm quite surprised. You said "IMO", did you try ?

> my suggestion is to empty the table name if it is following the [default naming scheme](https://github.com/Combodo/iTop/blob/cd8dd596416d502a150dcfe0f95ab202ef9c0cb2/synchro/synchrodatasource.class.inc.php#L908-L928)...

That would be a good idea...
A solution might be to store the generated table name on datasource creation, so that dump_task can see it and check if the current table name is the same ? Maybe I need to have a chat about this with @etaloc

piRGoif:
>  Maybe I need to have a chat about this with @etaloc

Done last week.
But I'll need to check with more people in Combodo, will organize a meeting.

piRGoif:
Modification validated after an internal meeting today.

PR #38, opened 2023-09-08
N°6773 Log the synchro error message in debug mode

Hipska:
It was hard to find out the actual error message returned when this is the only message:

> Synchronization of data source 'xxxx' failed.

So now if you set `console_log_level` to debug, you will also see the actual returned message by `synchro_exec.php`.

### DIFF
--- a/core/orchestrator.class.inc.php
+++ b/core/orchestrator.class.inc.php
@@ -264,7 +264,6 @@ class Orchestrator
 			Utils::SetCollector($oCollector, "Synchronize");
 			$bResult = $oCollector->Synchronize();
 			if (!$bResult) {
-				Utils::Log(LOG_DEBUG, $oCollector->GetErrorMessage());
 				if ($bStopOnError) {
 					break;
 				} else {
--- a/core/orchestrator.class.inc.php
+++ b/core/orchestrator.class.inc.php
@@ -264,6 +264,7 @@ class Orchestrator
 			Utils::SetCollector($oCollector, "Synchronize");
 			$bResult = $oCollector->Synchronize();
 			if (!$bResult) {
+				Utils::Log(LOG_DEBUG, $oCollector->GetErrorMessage());
 				if ($bStopOnError) {
 					break;
 				} else {
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -733,6 +733,7 @@ abstract class Collector
 				}
 			} else {
 				Utils::Log(LOG_ERR, "Synchronization of data source '{$this->sSourceName}' failed.");
+				Utils::Log(LOG_DEBUG, $sResult);
 				$this->sErrorMessage .= $sResult;
 				$iErrorsCount = 1;
 			}


piRGoif (approved):

piRGoif:
Setting to be reviewed in functional review as it will happen next week. If accepted, will need to be checked in technical review

piRGoif:
Functional review : ok, but :
- check if there is an existing bug ? 
- check with Olivier if he did something to fix it, or similar ?

piRGoif:
Seems like Saas team did something with N°4364, awaiting Olivier feedback

core/orchestrator.class.inc.php
```diff
@@ -264,6 +264,7 @@ public function Synchronize($aCollectors)
 			Utils::SetCollector($oCollector, "Synchronize");
 			$bResult = $oCollector->Synchronize();
 			if (!$bResult) {
+				Utils::Log(LOG_DEBUG, $oCollector->GetErrorMessage());
```
odain-cbd, line 267: this is a good idea for supportability. but why dont you aggregate synchro error message in the case of "xxx synchro failed" message instead?

cf collector.inc.class.php:
```suggestion
  Utils::Log(LOG_ERR, "Synchronization of data source '{$this->sSourceName}' failed: " . $sResult);
```

that way log message will have more detailed information in ERROR level. which would help those who configure collector to create event issue on iTop side for each ERROR/CRITICILAL log.

Hipska: I didn't want to interfere the current behaviour too much. I also thought this was an intended thing, as currently the error message wasn't used (`Collector::GetErrorMessage()` method was never called, so why store the result in `$sErrorMessage` anyway?)

So, you want me to change the existing error message to include the result from the API call? (this might also include a lot of html code in some situations)

odain-cbd: Hello,

I have thought about current PR and I have changed my mind.

the DEBUG log you intend to add will be redundant with some others in ERROR level generated in Collector::Synchronize().

it would be better to replace below line:
`
Utils::Log(LOG_ERR, "Synchronization of data source '{$this->sSourceName}' failed.");
$this->sErrorMessage .= $sResult;
`

by
`
Utils::Log(LOG_ERR, "Synchronization of data source '{$this->sSourceName}' failed. " . $sResult);
$this->sErrorMessage .= $sResult;
`

Hipska: Did you read my previous message?

odain-cbd: "Did you read my previous message?"
yes

"why store the result in $sErrorMessage anyway?)"
collector-base is an API. on a real collector design level, you may need to use the synchro output error. getter can seem useless in collector-base level. and still be called somewhere else.

"include a lot of html code in some situations"
good point. maybe it will be too much noise. in that case I suggest to keep your original idea and print that log in DEBUG. but only from Collector::Synchronize() context to avoid duplicate logs.

`
Utils::Log(LOG_ERR, "Synchronization of data source '{$this->sSourceName}' failed. "); 
Utils::Log(LOG_DEBUG, "$sResult");
`

Hipska: Okay, but I don't get what you mean with duplicate logs.. Where do you see anything duplicated?

odain-cbd: most of the time Collector::Synchronize() will log the issue in ERROR level AND store details (usually same message) in $this->sErrorMessage). that will come twice if you keep the new DEBUG log in Orchestrator::Synchronize().

cf examples of log that will become redundant
- "Failed to login to iTop. Invalid (or insufficent) credentials."
- "Synchronization of data source '{$this->sSourceName}' answered: {$aMatches[0][$idx]}"

I also added another case in https://github.com/Combodo/itop-data-collector-base/pull/42
- "Synchronization of data source '{$this->sSourceName}' answered: $idx"

except in your case where we forgot to log the details. that is the issue you intend to fix.
"Synchronization of data source '{$this->sSourceName}' failed."

Hipska: Right, looked over that.

Applied in 01b42985260e0dd4eb4f685fdbe85b8f323b4223.

Molkobain (approved):

PR #39, opened 2023-09-19
N°7506 - Add method to check if a specific module is installed in iTop

Hipska:
I wasn't sure to which class it could be added. Might as well be a static method on `Utils` or even somewhere else.

### DIFF
--- a/core/restclient.class.inc.php
+++ b/core/restclient.class.inc.php
@@ -225,7 +225,7 @@ class RestClient
 		try {
 			if (!isset($this->sLastInstallDate)) {
 				$aDatamodelResults = static::Get('ModuleInstallation', ['name' => 'datamodel'], 'installed', 1);
-				if ($aDatamodelResults['code'] != 0 || count($aDatamodelResults['objects']) === 0){
+				if ($aDatamodelResults['code'] != 0 || empty($aDatamodelResults['objects'])){
 					throw new Exception($aDatamodelResults['message'], $aDatamodelResults['code']);
 				}
 				$aDatamodel = current($aDatamodelResults['objects']);
@@ -233,7 +233,7 @@ class RestClient
 			}
 			
 			$aResults = static::Get('ModuleInstallation', ['name' => $sName, 'installed' => $this->sLastInstallDate], 'name,version', 1);
-			if ($aResults['code'] != 0 || count($aResults['objects']) === 0) {
+			if ($aResults['code'] != 0 || empty($aResults['objects'])) {
 				throw new Exception($aResults['message'], $aResults['code']);
 			}
 			$aObject = current($aResults['objects']);
--- a/core/restclient.class.inc.php
+++ b/core/restclient.class.inc.php
@@ -17,6 +17,7 @@
 class RestClient
 {
 	protected $sVersion;
+	protected $sLastInstallDate;
 
 	public function __construct()
 	{
@@ -34,13 +35,14 @@ class RestClient
 	}
 
 
-	public function Get($sClass, $keySpec, $sOutputFields = '*')
+	public function Get($sClass, $keySpec, $sOutputFields = '*', $iLimit = 0)
 	{
 		$aOperation = array(
 			'operation'     => 'core/get', // operation code
 			'class'         => $sClass,
 			'key'           => $keySpec,
 			'output_fields' => $sOutputFields, // list of fields to show in the results (* or a,b,c)
+			'limit'         => $iLimit,
 		);
 
 		return self::ExecOperation($aOperation, $this->sVersion);
@@ -215,7 +217,16 @@ class RestClient
 	public function CheckModuleInstallation(string $sName, bool $bRequired = false): bool
 	{
 		try {
-			$aResults = static::Get('ModuleInstallation', ['name' => $sName], 'name,version');
+			if (!isset($this->sLastInstallDate)) {
+				$aDatamodelResults = static::Get('ModuleInstallation', ['name' => 'datamodel'], 'installed', 1);
+				if ($aDatamodelResults['code'] != 0 || empty($aDatamodelResults['objects'])){
+					throw new Exception($aDatamodelResults['message'], $aDatamodelResults['code']);
+				}
+				$aDatamodel = current($aDatamodelResults['objects']);
+				$this->sLastInstallDate = $aDatamodel['fields']['installed'];
+			}
+			
+			$aResults = static::Get('ModuleInstallation', ['name' => $sName, 'installed' => $this->sLastInstallDate], 'name,version', 1);
 			if ($aResults['code'] != 0 || empty($aResults['objects'])) {
 				throw new Exception($aResults['message'], $aResults['code']);
 			}
--- a/core/restclient.class.inc.php
+++ b/core/restclient.class.inc.php
@@ -212,8 +212,8 @@ class RestClient
 	}
 	
 	/**
-	 * Check if the given module is installed in iTop
-	 * Mind that this assume that ModuleInstallation class is ordered by descending installation date
+	 * Check if the given module is installed in iTop.
+	 * Mind that this assumes the `ModuleInstallation` class is ordered by descending installation date
 	 *
 	 * @param string $sName Name of the module to be found
 	 * @param bool $bRequired Whether to throw exceptions when module not found
@@ -242,8 +242,7 @@ class RestClient
 			$sMessage = sprintf('%s iTop module %s is considered as not installed due to: %s', $bRequired ? 'Required' : 'Optional', $sName, $e->getMessage());
 			if ($bRequired) {
 				throw new Exception($sMessage, 0, $e);
-			}
-			else {
+			} else {
 				Utils::Log(LOG_INFO, $sMessage);
 				return false;
 			}
--- a/core/restclient.class.inc.php
+++ b/core/restclient.class.inc.php
@@ -204,4 +204,31 @@ class RestClient
 
 		return $bResult;
 	}
+	
+	/**
+	 * Check if the given module is installed in iTop
+	 * @param string $sName Name of the module to be found
+	 * @param bool $bRequired Whether to throw exceptions when module not found
+	 * @return bool True when the given module is installed, false otherwise
+	 * @throws Exception When the module is required but could not be found
+	 */
+	public function CheckModuleInstallation(string $sName, bool $bRequired = false): bool
+	{
+		try {
+			$aResults = static::Get('ModuleInstallation', ['name' => $sName], 'name,version');
+			if ($aResults['code'] != 0 || empty($aResults['objects'])) {
+				throw new Exception($aResults['message'], $aResults['code']);
+			}
+			$aObject = current($aResults['objects']);
+			Utils::Log(LOG_DEBUG, sprintf('iTop module %s version %s is installed.', $aObject['fields']['name'], $aObject['fields']['version']));
+		} catch (Exception $e) {
+			$sMessage = sprintf('%s iTop module %s is considered as not installed due to: %s', $bRequired ? 'Required' : 'Optional', $sName, $e->getMessage());
+			if ($bRequired) throw new Exception($sMessage, 0, $e);
+			else {
+				Utils::Log(LOG_INFO, $sMessage);
+				return false;
+			}
+		}
+		return true;
+	}
 }
--- a/core/restclient.class.inc.php
+++ b/core/restclient.class.inc.php
@@ -17,11 +17,6 @@
 class RestClient
 {
 	protected $sVersion;
-	/**
-	 * @var string Keeps track of the latest date the datamodel has been installed/updated
-	 * (in order to check which modules were installed with it)
-	 */
-	protected $sLastInstallDate;
 
 	public function __construct()
 	{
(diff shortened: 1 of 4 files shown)


Molkobain (changes requested):
Discussed with Denis, as-is the PR doesn't suit us for the following reasons:
  * Just looking for a module in that class / table can return true even though the module has been uninstalled. This table is an history, not the current state of the instance.
  * Ideally, it would look for the most recent "datamodel %" entry, than look for the module at the exact same datetime.

Hipska:
Oh, that's right indeed. I'll see how it could be retrieved the best then...

Molkobain:
It seems that the `ModuleInstallation` class is already ordered by desc. installation datetime ([see here](https://github.com/Combodo/iTop/blob/develop/setup/moduleinstallation.class.inc.php#L42)), so you could:
  * Make a first call to `core/get`, with a limit to `1` for `SELECT ModuleInstallation WHERE name = "datamodel"`, which will return the last compilation datetime.
    * Mind to make a cache for that datetime info, so you don't have to make the call again if you are checking more than 1 module.
  * Then make a second call to `core/get`, also with a limit to `1` for `SELECT ModuleInstallation WHERE name = "<YOUR_MODULE_CODE>" AND installed = "<DATETIME_FROM_PREVIOUS_CALL>"`

Hipska:
Yeah, also had something like this in mind. Please see 8b8849d.

core/restclient.class.inc.php
```diff
@@ -204,4 +206,40 @@ public static function GetFullSynchroDataSource(&$aSource, $iSourceId)
...
+	 * @param bool $bRequired Whether to throw exceptions when module not found
+	 * @return bool True when the given module is installed, false otherwise
+	 * @throws Exception When the module is required but could not be found
+	 */
+	public function CheckModuleInstallation(string $sName, bool $bRequired = false): bool
+	{
+		try {
+			if (!isset($this->sLastInstallDate)) {
+				$aDatamodelResults = static::Get('ModuleInstallation', ['name' => 'datamodel'], 'installed', 1);
+				if ($aDatamodelResults['code'] != 0 || empty($aDatamodelResults['objects'])){
```
Molkobain, line 226:
```suggestion
				if ($aDatamodelResults['code'] != 0 || count($aDatamodelResults['objects']) === 0){
```

We tend to avoid using `empty()`

core/restclient.class.inc.php
```diff
@@ -204,4 +206,40 @@ public static function GetFullSynchroDataSource(&$aSource, $iSourceId)
 
 		return $bResult;
 	}
+	
+	/**
+	 * Check if the given module is installed in iTop
```
Molkobain, line 215:
```suggestion
	 * Check if the given module is installed in iTop
	 * Mind that this assume that ModuleInstallation class is ordered by descending installation date
	 *
```

core/restclient.class.inc.php
```diff
@@ -204,4 +206,40 @@ public static function GetFullSynchroDataSource(&$aSource, $iSourceId)
...
+			
+			$aResults = static::Get('ModuleInstallation', ['name' => $sName, 'installed' => $this->sLastInstallDate], 'name,version', 1);
+			if ($aResults['code'] != 0 || empty($aResults['objects'])) {
+				throw new Exception($aResults['message'], $aResults['code']);
+			}
+			$aObject = current($aResults['objects']);
+			Utils::Log(LOG_DEBUG, sprintf('iTop module %s version %s is installed.', $aObject['fields']['name'], $aObject['fields']['version']));
+		} catch (Exception $e) {
+			$sMessage = sprintf('%s iTop module %s is considered as not installed due to: %s', $bRequired ? 'Required' : 'Optional', $sName, $e->getMessage());
+			if ($bRequired) throw new Exception($sMessage, 0, $e);
```
Molkobain, line 241:
```suggestion
			if ($bRequired) {
				throw new Exception($sMessage, 0, $e);
			}
```

We rather not use inline condition

Molkobain (changes requested):
Can you add a unit test that check:
  * That a common module is present (eg. `itop-structure` as tests run on an iTop 3.2, or `itop-config`)
  * That a fake module is not present (eg. `fake-module`)

core/restclient.class.inc.php
```diff
@@ -204,4 +210,40 @@ public static function GetFullSynchroDataSource(&$aSource, $iSourceId)
...
+				$aDatamodelResults = static::Get('ModuleInstallation', ['name' => 'datamodel'], 'installed', 1);
+				if ($aDatamodelResults['code'] != 0 || empty($aDatamodelResults['objects'])){
+					throw new Exception($aDatamodelResults['message'], $aDatamodelResults['code']);
+				}
+				$aDatamodel = current($aDatamodelResults['objects']);
+				$this->sLastInstallDate = $aDatamodel['fields']['installed'];
+			}
+			
+			$aResults = static::Get('ModuleInstallation', ['name' => $sName, 'installed' => $this->sLastInstallDate], 'name,version', 1);
+			if ($aResults['code'] != 0 || empty($aResults['objects'])) {
```
Hipska, line 234:
```suggestion
			if ($aResults['code'] != 0 || count($aResults['objects']) === 0) {
```

core/restclient.class.inc.php
Hipska: I'll revert this since this will give warnings when no objects returned (datatype is `null`). See https://github.com/Combodo/iTop/pull/231

Hipska:
I lost a full day trying to get the tests working, so I added what I got and should work, but the old version of phpunit does not seem to do what I'm expecting it to do somehow..

core/restclient.class.inc.php
Molkobain: Good catch

Molkobain (approved):

Molkobain:
Thank you very much for coding the extra mile with the unit test, it means a lot.

Molkobain:
_Just discussing the unit test with @odain-cbd before merging_

odain-cbd:
Dear Thomas,

I just cloned your fork and indeed there issues with PHPUNIT somehow. we will upgrade this lib.

test/RestTest.php
```diff
@@ -116,4 +116,27 @@ function($sKey, $aDefaultValue) use ($aParameters) {
 		$this->assertEquals(['retcode' => 0], $oRestClient->ListOperations());
 	}
```
odain-cbd, line 118: I suggest to move this function somewhere else to mock RestClient and be able to test this new feature. for example we could do it in Utils in a static call (not the best solution). here is a possible way:
`
	/**
	 * @var string Keeps track of the latest date the datamodel has been installed/updated
	 * (in order to check which modules were installed with it)
	 */
	protected static $sLastInstallDate;

	/**
	 * Check if the given module is installed in iTop.
	 * Mind that this assumes the `ModuleInstallation` class is ordered by descending installation date
	 *
	 * @param string $sName Name of the module to be found
	 * @param bool $bRequired Whether to throw exceptions when module not found
	 * @param \RestClient|null $oClient: provider only in test conditions
	 *
	 * @return bool True when the given module is installed, false otherwise
	 * @throws Exception When the module is required but could not be found
	 */
	public static function CheckModuleInstallation(string $sName, bool $bRequired = false, RestClient $oClient = null): bool
	{
		if (is_null($oClient)){
			$oClient = new RestClient();
		}

		try {
			if (!isset(static::$sLastInstallDate)) {
				$aDatamodelResults = $oClient->Get('ModuleInstallation', ['name' => 'datamodel'], 'installed', 1);
				if ($aDatamodelResults['code'] != 0 || empty($aDatamodelResults['objects'])){
					throw new Exception($aDatamodelResults['message'], $aDatamodelResults['code']);
				}
				$aDatamodel = current($aDatamodelResults['objects']);
				static::$sLastInstallDate = $aDatamodel['fields']['installed'];
			}

			$aResults = $oClient->Get('ModuleInstallation', ['name' => $sName, 'installed' => static::$sLastInstallDate], 'name,version', 1);
			if ($aResults['code'] != 0 || empty($aResults['objects'])) {
				throw new Exception($aResults['message'], $aResults['code']);
			}
			$aObject = current($aResults['objects']);
			Utils::Log(LOG_DEBUG, sprintf('iTop module %s version %s is installed.', $aObject['fields']['name'], $aObject['fields']['version']));
		} catch (Exception $e) {
			$sMessage = sprintf('%s iTop module %s is considered as not installed due to: %s', $bRequired ? 'Required' : 'Optional', $sName, $e->getMessage());
			if ($bRequired) {
				throw new Exception($sMessage, 0, $e);
			} else {
				Utils::Log(LOG_INFO, $sMessage);
				return false;
			}
		}
		return true;
	}
`

core/restclient.class.inc.php
```diff
@@ -204,4 +210,43 @@ public static function GetFullSynchroDataSource(&$aSource, $iSourceId)
 
 		return $bResult;
 	}
+
```
odain-cbd, line 213: I suggest to move this function somewhere else to mock RestClient and be able to test this new feature. for example we could do it in Utils in a static call (not the best solution). here is a possible way to illustrate how to mock and test it:

`

/**
 * @var string Keeps track of the latest date the datamodel has been installed/updated
 * (in order to check which modules were installed with it)
 */
protected static $sLastInstallDate;

/**
 * Check if the given module is installed in iTop.
 * Mind that this assumes the `ModuleInstallation` class is ordered by descending installation date
 *
 * @param string $sName Name of the module to be found
 * @param bool $bRequired Whether to throw exceptions when module not found
 * @param \RestClient|null $oClient: provider only in test conditions
 *
 * @return bool True when the given module is installed, false otherwise
 * @throws Exception When the module is required but could not be found
 */
public static function CheckModuleInstallation(string $sName, bool $bRequired = false, RestClient $oClient = null): bool
{
	if (is_null($oClient)){
		$oClient = new RestClient();
	}

	try {
		if (!isset(static::$sLastInstallDate)) {
			$aDatamodelResults = $oClient->Get('ModuleInstallation', ['name' => 'datamodel'], 'installed', 1);
			if ($aDatamodelResults['code'] != 0 || empty($aDatamodelResults['objects'])){
				throw new Exception($aDatamodelResults['message'], $aDatamodelResults['code']);
			}
			$aDatamodel = current($aDatamodelResults['objects']);
			static::$sLastInstallDate = $aDatamodel['fields']['installed'];
		}

		$aResults = $oClient->Get('ModuleInstallation', ['name' => $sName, 'installed' => static::$sLastInstallDate], 'name,version', 1);
		if ($aResults['code'] != 0 || empty($aResults['objects'])) {
			throw new Exception($aResults['message'], $aResults['code']);
		}
		$aObject = current($aResults['objects']);
		Utils::Log(LOG_DEBUG, sprintf('iTop module %s version %s is installed.', $aObject['fields']['name'], $aObject['fields']['version']));
	} catch (Exception $e) {
		$sMessage = sprintf('%s iTop module %s is considered as not installed due to: %s', $bRequired ? 'Required' : 'Optional', $sName, $e->getMessage());
		if ($bRequired) {
			throw new Exception($sMessage, 0, $e);
		} else {
			Utils::Log(LOG_INFO, $sMessage);
			return false;
		}
	}
	return true;
}

`

test/RestTest.php
odain-cbd, line 118: here is my suggestion to test your logic and not the mocks:

```
public function testCheckModuleInstallation(){
		$oRestClient = $this->createMock(\RestClient::class);

		$oReflectionLastInstallDate = new \ReflectionProperty(\Utils::class, 'sLastInstallDate');
		$oReflectionLastInstallDate->setAccessible(true);
		$oReflectionLastInstallDate->setValue($oRestClient, '0000-00-00 00:00:00');

		$oRestClient->expects($this->exactly(2))
			->method('Get')
			->willReturnMap([
				['ModuleInstallation', ['name' => 'itop-structure', 'installed' => '0000-00-00 00:00:00'], 'name,version', 1, [
					'code' => 0,
					'objects' => ['ModuleInstallation::0' => ['fields' => ['name' => 'itop-structure', 'version' => '0.0.0']]],
				]],
				['ModuleInstallation', ['name' => 'fake-module', 'installed' => '0000-00-00 00:00:00'], 'name,version', 1, [
					'code' => 0,
					'objects' => null,
				]],
			]);

		$this->assertTrue(Utils::CheckModuleInstallation('itop-structure', true, $oRestClient));
		$this->assertFalse(Utils::CheckModuleInstallation('fake-module', false, $oRestClient));
	}
```

odain-cbd (changes requested):
please see my proposal

Hipska:
I don't feel like this method fits well into the Utils class, but I have moved and fixed the test.

Molkobain:
Discussing this with Stephen, wouldn't it be better if we could pass an optional "min version" for the module?
It could be helpful to adapt what we send when there is a DM change in a module version.

odain-cbd:
"I don't feel like this method fits well into the Utils class, but I have moved and fixed the test."

I agree with you. maybe it should be located somewhere else than Utils. I just proposed it to let you validate this feature. now that is tested you may easily move it in a proper place. maybe a new class?

Hipska:
> Discussing this with Stephen, wouldn't it be better if we could pass an optional "min version" for the module? It could be helpful to adapt what we send when there is a DM change in a module version.

@Molkobain I'm not sure if this should also be in the same method.. What do you propose as its prototype then? An optional argument or the method returning the latest version number (as it is now putting into the logs)?

I don't think the versions are that relevant for the collectors. The one I know that has module checks, only checks if the classes are available: [example](https://github.com/TeemIp/teemip-ip-discovery-collector/blob/546cdc1b76b3a0bd5cc87b374d511c46e8d7db00/src/TeemIpDiscoveryCollectionPlan.class.inc.php#L26-L121). For [my own use-case](https://github.com/Super-Visions/snmp-discovery-collector/blob/6da27a49758876aeb4c758ee560396182a9b8f7c/src/SnmpDiscoveryCollector.class.inc.php#L29-L30) I wouldn't mind the version as well. Do you have other examples?

> I agree with you. maybe it should be located somewhere else than Utils.

@odain-cbd I still feel it fits well in `RestClient` similar to `GetFullSynchroDataSource` also is located there. Maybe it could be a method of `Orchestrator`? Similar to `AddRequirements` and `CheckRequirements`, but those are related to PHP modules.

Hipska:
I was more asking for current existing collectors that actually check for a version or have a need for it. Your example is also a valid one.

I'm more in favour of this method returning the installed version, so the version can be checked elsewhere.

core/utils.class.inc.php
```diff
@@ -676,19 +676,25 @@ public static function ComputeCurlOptions(array $aRawCurlOptions, int $iCurrentT
 	 * Check if the given module is installed in iTop.
 	 * Mind that this assumes the `ModuleInstallation` class is ordered by descending installation date
 	 *
-	 * @param string $sName Name of the module to be found
+	 * @param string $sModuleId Name of the module to be found, optionally included version
```
Molkobain, line 679:
```suggestion
	 * @param string $sModuleId Name of the module to be found, optionally included version (e.g. "some-module" or "some-module/1.2.3")
```

test/UtilsTest.php
```diff
@@ -254,7 +254,9 @@ public function testCheckModuleInstallation(){
 				]],
 			]);
 		
-		$this->assertTrue(Utils::CheckModuleInstallation('itop-structure', true, $oRestClient));
+		$this->assertTrue(Utils::CheckModuleInstallation('itop-structure', false, $oRestClient));
+		$this->assertTrue(Utils::CheckModuleInstallation('itop-structure/0.0.0', false, $oRestClient));
+		$this->assertFalse(Utils::CheckModuleInstallation('itop-structure/1.2.3', false, $oRestClient));
```
Molkobain, line 259: I'm not sure to understand this test case, `itop-structure` was introduced in 3.0 with first version being `itop-structure/3.0.0`, so how can the test for "0.0.0" return true, but not the test for "1.2.3"?

Molkobain (approved):

odain-cbd (approved):

PR #41, opened 2023-09-21
N°6956 - Improve \CollectionPlan PHPDoc to ease DX in IDE

Hipska:
In order to let IDE get better instructions

### DIFF
--- a/core/collectionplan.class.inc.php
+++ b/core/collectionplan.class.inc.php
@@ -40,9 +40,9 @@ abstract class CollectionPlan
 	}
 
 	/**
-	 * @return \CollectionPlan
+	 * @return static
 	 */
-	public static function GetPlan(): CollectionPlan
+	public static function GetPlan(): static
 	{
 		return self::$oCollectionPlan;
 	}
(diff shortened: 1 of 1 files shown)


piRGoif (approved):

core/collectionplan.class.inc.php
```diff
@@ -40,17 +40,17 @@ public function Init(): void
 	}
 
 	/**
-	 * @return \CollectionPlan
+	 * @return static
 	 */
-	public static function GetPlan(): CollectionPlan
+	public static function GetPlan(): static
```
Molkobain, line 45: This is not supported by PHP 7.x, only PHP 8.0+

One solution could be to remove the return type hint completely and keep the PHPDoc as you suggested. That's what we did in iTop 3.0 for UIBlock for example.

Molkobain (changes requested):

core/collectionplan.class.inc.php
Hipska: That's right, will change to this:
```suggestion
	public static function GetPlan()
```

Molkobain (approved):

core/collectionplan.class.inc.php
piRGoif: Woops, good to know, good catch Guillaume, thanks !

Molkobain:

Molkobain:
Accepted during functional review

PR #43, opened 2023-12-22
N°7338 - Use "itop_login" as fallback value if no "synchro_user" defined in parameters

Hipska:
Add useful default value for `synchro_user` so there is no need to define the same username in most cases.


Molkobain (approved):

Molkobain:

Molkobain:
Accepted during functional review once comment is updated. (Sorry I can't commit on this fork)

PR #44, opened 2024-01-15
JsonCollector: Introduce GetOptionalHeaders() and GetCurlOptions() methods

jbostoen:
This PR intends to make it easier to extend the JsonCollector.

**Use case:** A third-party API which requires Basic Authorization with each POST request. 

Basic Authorization can basically be performed in these ways:

1. Using a cURL option: `CURLOPT_USERPWD` . 
2. By explicitly adding it in the HTTP headers of the POST request.

**Without the PR:**

- cURL option: Now, this would however need to be defined in the generic "params.xxx.xml" ("conf" dir). However, most third-party application data seems to be set in the "params.xxx.xml" file of the "collectors" dir.
- To set custom headers, the whole Prepare() method would need to be edited.

**WIth the PR:**

- cURL options can also be defined in the "collectors" XML file.
- It's also easy to extend the JsonCollector and add a custom HTTP header. The authentication could be coded instead depending on a separate "user" and "pwd" setting in the collector's XML.

By default, no HTTP headers are set.
By default, cURL options = cURL options as defined in the "conf" XML (as before); and cURL options as defined in the "collectors" XML (these are superseding).

Note: I used the correct spelling of "optional". In the source code, it's consistently written incorrectly with "optionnal". I was considering updating it, but that might cause issues if people already extended classes and if those methods were overridden.

### DIFF
--- a/core/jsoncollector.class.inc.php
+++ b/core/jsoncollector.class.inc.php
@@ -141,12 +141,13 @@ abstract class JsonCollector extends Collector
 			} else {
 				$aDataGet = [];
 			}
-			$iSynchroTimeout = (int)Utils::GetConfigurationValue('itop_synchro_timeout', 600); // timeout in seconds, for a synchro to run
-			$aCurlOptions = Utils::GetCurlOptions($iSynchroTimeout);
+			
+			$aCurlOptions = static::GetCurlOptions();
+			
 
 			//logs
 			Utils::Log(LOG_DEBUG, 'Source aDataGet: '.json_encode($aDataGet));
-			$this->sFileJson = Utils::DoPostRequest($this->sURL, $aDataGet, '', $aResponseHeaders, $aCurlOptions);
+			$this->sFileJson = Utils::DoPostRequest($this->sURL, $aDataGet, static::GetOptionalHeaders(), $aResponseHeaders, $aCurlOptions);
 			Utils::Log(LOG_DEBUG, 'Source sFileJson: '.$this->sFileJson);
 			Utils::Log(LOG_INFO, 'Synchro URL (target): '.Utils::GetConfigurationValue('itop_url', array()));
 		} else {
@@ -257,6 +258,51 @@ abstract class JsonCollector extends Collector
 		}
 
 		return false;
+	}
+	
+	/**
+	 * Returns optional HTTP headers which can be sent to the specified URL. 
+	 * For example, a subclass can be override this method to send an Authorization: header.
+	 */
+	public function GetOptionalHeaders() : string
+	{
+		
+		return '';
+		
+	}
+	
+	/**
+	 * Returns cURL options to use.
+	 * For example, a subclass can be override this method to add the CURLOPT_USERPWD and a value.
+	 */
+	public function GetCurlOptions() : array
+	{
+		
+		$iSynchroTimeout = (int)Utils::GetConfigurationValue('itop_synchro_timeout', 600); // timeout in seconds, for a synchro to run
+		$aParamsSourceJson = $this->aCollectorConfig;
+
+		// Step 1:
+		// Use cURL options from "config" XML.
+		$aConfigCurlOptions = Utils::GetCurlOptions($iSynchroTimeout);
+		
+		// Step 2:
+		// If available: Add or update options defined in the "collector" XML.
+		// Apply the same mechanism to convert constants. The flexibility in place is to allow adding and overriding values, so do not pass on synchro timeout.
+		$aCollectorCurlOptions = array();
+
+		if(isset($aParamsSourceJson['curl_options'])) {	
+			$aCollectorCurlOptions = $aParamsSourceJson['curl_options'];
+		}
+		if(isset($aParamsSourceJson['CURL_OPTIONS'])) {	
+			$aCollectorCurlOptions = $aParamsSourceJson['CURL_OPTIONS'];
+		}
+		
+		if(count($aCollectorCurlOptions) > 0) {
+			$aCollectorCurlOptions = utils::GetCurlOptions($aCollectorCurlOptions, -1);
+		}
+		
+		return array_replace($aConfigCurlOptions, $aCollectorCurlOptions);
+			
 	}
 
 	/**


Molkobain (changes requested):

core/jsoncollector.class.inc.php
```diff
@@ -258,6 +259,51 @@ public function Fetch()
...
+		// If available: Add or update options defined in the "collector" XML.
+		// Apply the same mechanism to convert constants. The flexibility in place is to allow adding and overriding values, so do not pass on synchro timeout.
+		$aCollectorCurlOptions = array();
+
+		if(isset($aParamsSourceJson['curl_options'])) {	
+			$aCollectorCurlOptions = $aParamsSourceJson['curl_options'];
+		}
+		if(isset($aParamsSourceJson['CURL_OPTIONS'])) {	
+			$aCollectorCurlOptions = $aParamsSourceJson['CURL_OPTIONS'];
+		}
```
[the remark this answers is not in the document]
jbostoen: I thought so as well - but then again, I'm trying to be consistent: see https://github.com/Combodo/itop-data-collector-base/blob/34e2c03d0dc6c9b54b54ad93ae5bac3295e1b982/core/jsoncollector.class.inc.php#L72 and other examples in the JsonCollector. Not sure if the uppercase version is for legacy or has a different purpose.

core/jsoncollector.class.inc.php
```diff
@@ -258,6 +259,47 @@ public function Fetch()
...
+		
+		if(count($aCollectorCurlOptions) > 0) {
+			$aCollectorCurlOptions = utils::GetCurlOptions($aCollectorCurlOptions, -1);
```
Hipska, line 298:
```suggestion
			$aCollectorCurlOptions = Utils::GetCurlOptions($aCollectorCurlOptions, -1);
```
What is the purpose of this section anyway?

core/jsoncollector.class.inc.php
```diff
@@ -141,12 +141,13 @@ public function Prepare()
 			} else {
 				$aDataGet = [];
 			}
-			$iSynchroTimeout = (int)Utils::GetConfigurationValue('itop_synchro_timeout', 600); // timeout in seconds, for a synchro to run
-			$aCurlOptions = Utils::GetCurlOptions($iSynchroTimeout);
+			
+			$aCurlOptions = static::GetCurlOptions();
```
Hipska, line 145: Cannot be called static
```suggestion
			$aCurlOptions = $this->GetCurlOptions();
```

core/jsoncollector.class.inc.php
```diff
@@ -141,12 +141,13 @@ public function Prepare()
...
+			
 
 			//logs
 			Utils::Log(LOG_DEBUG, 'Source aDataGet: '.json_encode($aDataGet));
-			$this->sFileJson = Utils::DoPostRequest($this->sURL, $aDataGet, '', $aResponseHeaders, $aCurlOptions);
+			$this->sFileJson = Utils::DoPostRequest($this->sURL, $aDataGet, static::GetOptionalHeaders(), $aResponseHeaders, $aCurlOptions);
```
Hipska, line 150: I would suggest to not call it statically, but with the object instance:
```suggestion
			$this->sFileJson = Utils::DoPostRequest($this->sURL, $aDataGet, $this->GetOptionalHeaders(), $aResponseHeaders, $aCurlOptions);
```

Hipska (changes requested):

core/jsoncollector.class.inc.php
```diff
@@ -258,6 +259,47 @@ public function Fetch()
 
 		return false;
 	}
+	
+	/**
+	 * @return Optional HTTP headers which can be sent to the specified URL.
```
Molkobain, line 264:
```suggestion
	 * @return Optional HTTP headers which can be sent to the specified URL. {@see \Utils::DoPostRequest() $sOptionnalHeaders parameter for the expected synthax}
```

core/jsoncollector.class.inc.php
```diff
@@ -258,6 +259,51 @@ public function Fetch()
 
 		return false;
 	}
+	
+	/**
+	 * Returns optional HTTP headers which can be sent to the specified URL. 
+	 * For example, a subclass can be override this method to send an Authorization: header.
```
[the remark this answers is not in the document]
Molkobain: I just added a suggestion above, feel free to adjust it.

core/jsoncollector.class.inc.php
```diff
@@ -258,6 +259,47 @@ public function Fetch()
...
+		if(isset($aParamsSourceJson['curl_options'])) {	
+			$aCollectorCurlOptions = $aParamsSourceJson['curl_options'];
+		}
+		if(isset($aParamsSourceJson['CURL_OPTIONS'])) {	
+			$aCollectorCurlOptions = $aParamsSourceJson['CURL_OPTIONS'];
+		}
+		
+		if(count($aCollectorCurlOptions) > 0) {
+			$aCollectorCurlOptions = Utils::GetCurlOptions($aCollectorCurlOptions, -1);
+		}
```
Hipska, line 299: What is the purpose of this? Doesn't seem a correct call to GetCurlOptions.

core/jsoncollector.class.inc.php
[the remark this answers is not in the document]
Molkobain: Indeed... I was not aware of that 😅

core/jsoncollector.class.inc.php
jbostoen: My bad in the draft, should have been the one which converts the options.

Hipska: Oh I see. It doesn't need the `if` anyway.

Hipska (approved):

odain-cbd:
I come after all other feedbacks. everything is ok to me. great job and idea!

NB:
today that part of the code was not designed to easily test it. obviously. so i will not push for tests. we will discuss it apart from this PR. there is much more refactoring to reach that goal...

odain-cbd (approved):

Molkobain (approved):

Molkobain:
Technical review:
  * Logic seems ok
  * Guillaume will just check if the refacto can be done on a prent class for a better genericity or if it's meant to be only in JSON collector and children.

PR #45, opened 2024-02-09
N°3709 - Add the possibility to use an other date format than the default one

accognet:
By adding 
`  <date_format>d/m/Y</date_format>`
in <parameters> tag

### DIFF
--- a/conf/params.distrib.xml
+++ b/conf/params.distrib.xml
@@ -106,4 +106,5 @@
   <contact_to_notify></contact_to_notify>
   <!-- iTop user set as allowed to run synchronization. It is highly recommended to use the same as itop_login -->
   <synchro_user></synchro_user>
+  <date_format>Y-m-d</date_format>
 </parameters>
--- a/conf/params.distrib.xml
+++ b/conf/params.distrib.xml
@@ -106,5 +106,6 @@
   <contact_to_notify></contact_to_notify>
   <!-- iTop user set as allowed to run synchronization. It is highly recommended to use the same as itop_login -->
   <synchro_user></synchro_user>
+  <!-- date format in collected data -->
   <date_format>Y-m-d</date_format>
 </parameters>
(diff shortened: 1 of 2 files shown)


Hipska (approved):

test/CollectorSynchroTest.php
```diff
@@ -63,7 +63,8 @@ public function testSynchroOutput($iConsoleLogLevel, $sExpectedOutputRequiredToI
 		    'no_stop_on_import_error' => 1,
 		    'output' => $sExpectedOutputRequiredToItopSynchro,
 		    'csvdata' => 'FAKECSVCONTENT',
-		    'charset' => 'UTF-8'
+		    'charset' => 'UTF-8',
+            'date_format' => 'Y-m-d'
```
Molkobain, line 67:
```suggestion
            'date_format' => 'd/m/Y'
```

test/utils/params.test.xml
```diff
@@ -19,5 +19,6 @@
   </curl_options>
   <contact_to_notify></contact_to_notify>
   <synchro_user></synchro_user>
+  <date_format>Y-m-d</date_format>
```
Molkobain, line 22:
```suggestion
  <date_format>d/m/Y</date_format>
```

Molkobain (changes requested):
Test case should not test the default value as it's not relevant.
Actually, test fails if we set a different value.

Molkobain (approved):
Test fixed with https://github.com/Combodo/itop-data-collector-base/pull/45/commits/ea56acb22f1a19b2004425be6acbd859e3a29d2a

Hipska:
@accognet @Molkobain Please be noted that this at least broke [teemip-ip-discovery-collector](https://github.com/TeemIp/teemip-ip-discovery-collector) and probably also others that sync dates and relied on the default behaviour.

The default in this base collector should be set to `Y-m-d H:i:s` as it is the default for `synchro_import.php`: https://github.com/Combodo/iTop/blob/36a6ee4fc967ad5f6ec9c57e94140efc865303d3/synchro/synchro_import.php#L91

PR #47, opened 2024-02-12
N°7339 - Improve error handling when user has no permission to write in data directory

Hipska:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | No

## Symptom
```
PHP Warning:  fopen(data/TeemIpDiscoveryIPv4Collector-1.csv): Failed to open stream: Permission denied in core/collector.class.inc.php on line 614
[2024-02-12 09:54:41]	[Error]	Unable to open the file 'data/TeemIpDiscoveryIPv4Collector-1.csv' for writing.
PHP Fatal error:  Uncaught TypeError: fputcsv(): Argument #1 ($stream) must be of type resource, bool given in core/collector.class.inc.php:589
Stack trace:
#0 core/collector.class.inc.php(589): fputcsv(false, Array, ';')
#1 core/collector.class.inc.php(555): Collector->AddHeader(Array)
#2 collectors/src/TeemIpDiscoveryIPv4Collector.class.inc.php(682): Collector->Collect('1000')
#3 core/orchestrator.class.inc.php(237): TeemIpDiscoveryIPv4Collector->Collect('1000', true)
#4 exec.php(123): Orchestrator->Collect(Array, '1000', true)
#5 {main}
  thrown in core/collector.class.inc.php on line 589
```

So the collector crashes.

## Reproduction procedure

1. With teemip-ip-discovery-collector 3.1.2 
2. With PHP 8.1.27 
3. Targeting TeemIP version 3.1.4-2311 
4. Run collector as a user without permission to write the collector data folder.

## Cause

Running the collector as a different user who doesn't have write permissions to the `data` directory and missing error handling.

## Proposed solution

Return an `IOException` instead of just logging the message, so it can be catched within the `Collector::Collect` method and shut down correctly.

### DIFF
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -603,26 +603,32 @@ abstract class Collector
 		fputcsv($this->aCSVFile[$this->iFileIndex], $aData, $this->sSeparator);
 	}
 
+	/**
+	 * @return true
+	 * @throws IOException When file cannot be opened.
+	 */
 	protected function OpenCSVFile()
 	{
-		$bResult = true;
 		if ($this->MustProcessBeforeSynchro()) {
 			$sDataFile = Utils::GetDataFilePath(get_class($this).'.raw-'.(1 + $this->iFileIndex).'.csv');
 		} else {
 			$sDataFile = Utils::GetDataFilePath(get_class($this).'-'.(1 + $this->iFileIndex).'.csv');
 		}
-		$this->aCSVFile[$this->iFileIndex] = fopen($sDataFile, 'wb');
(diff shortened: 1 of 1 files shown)


Molkobain:
This changes the current behavior, instead of just skipping the invalid file, it will stop the entire collector.
It will be discussed tomorrow during the functional review.

Hipska:
No, the current behaviour is not skipping, but crash of the collector. See the provided symptom (PHP Fatal error because of Uncaught TypeError)

Of course is fixing bugs change of behaviour. The question is do you really want the current behaviour?

Molkobain:
Accepted during functional review

PR #49, opened 2024-02-26
Fix lookuptable ignore errors

dw-itomig:
if the LookupTable set to "ignore-errors" the Lookup itself should not return false in this case. Otherwise the line will be ignored in any case if you follow the documentation [collector-base on iTophub](https://www.itophub.io/wiki/page?id=extensions:itop-data-collector-base#advanced_lookups). The Best-Practise should be to **throw New IgnoredRowException** if the lookup function return false AND if the ignore_errors was set to false for the LookupTable Object in general.
Please correct me if i am wrong.

Best Regards
David from ITOMIG

(If you implement this pull request - i think no configuration or side impact should be considered - just locally tested)


Molkobain (approved):

Molkobain:
Hello David, thanks for the PR!

PR #50, opened 2024-04-26
chore(requirements): Apply correct minimal versions

Hipska:
The current composer.json doesn't mach the versions supported according to the documentation, this makes it hard to test the collector code for supported versions higher than PHP 7.0.

Next to that, the version checker checks for PHP 5.3+, but since only PHP 7.x+ is supported, I changed the default version value.

I hope this PR description is clear enough, the questions/steps in the request template seemed not fitting for this kind of cleanup.


piRGoif:
Warning for PHPUnit the tests code isn't ready yet for versions > 6
There is a ticket somewhere in the Combodo bug tracker to update those tests.

Hipska:
It blocks new people to be able to install/use phpunit if I recall correctly.

Molkobain:
Yes please, remove it for now, so we can push this PR forward in the meantime.

composer.json
```diff
@@ -1,9 +1,9 @@
 {
    "require": {
-       "php": "7.*",
+       "php": "7.* || 8.0.* || 8.1.",
        "ext-libxml": "*"
    },
    "require-dev": {
-       "phpunit/phpunit": "^6"
+       "phpunit/phpunit": "*"
```
Hipska, line 7:
```suggestion
       "phpunit/phpunit": "^6"
```

Molkobain (approved):

jf-cbd:
Hello @Hipska, itop-data-collector-base extension is also compatible with PHP 8.2 and 8.3, could you please add them too so we can merge the PR ? Thanks a lot :)

composer.json
```diff
@@ -1,6 +1,6 @@
 {
    "require": {
-       "php": "7.*",
+       "php": "7.* || 8.0.* || 8.1.",
```
Hipska, line 3:
```suggestion
       "php": "7.* || 8.* <= 8.3",
```

jf-cbd:
Thanks for your contribution :)

PR #52, opened 2024-05-15
N°5139 - Let data that is null also be null at the other side-fix

odain-cbd:
Current PR intends to enhance "Reset or untouch" option (nullified configuration) to work also for empty string ("") fields values.

During data collection, fields can have a null value. during iTop object synchronization you can decide either 

- to reset the object field value on iTop side (during synchronization) and do nothing special in iTop configuration.
- to update the field to a default value configured on collector side. use "default" configuration section for that purpose
- to keep its previous value by leaving it untouch and sending '<NULL>' to iTop synchonization. you have to configure collector with a nullified section.

we propose to have same "untouch" behaviour for empty string values collected. as a matter of fact null cannot be collected for some type of collectors (like CSV). which make the "nullify" feature unusable... seems like a bug to me.

- cf documentation :

https://wiki.combodo.com/doku.php?id=extensions:itop-data-collector-base#reset_or_untouch

- cf PR that brought the feature:

https://github.com/Combodo/itop-data-collector-base/pull/7

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,6 +4,6 @@
        "ext-libxml": "*"
    },
    "require-dev": {
-       "phpunit/phpunit": "^6"
+     "phpunit/phpunit": "^9"
    }
 }


Hipska:
I'm only concerned on how to sync a NULL value into a nullable DECIMAL field in iTop. I think that is still not possible, even with these additions?

core/collector.class.inc.php
```diff
@@ -594,7 +594,7 @@ protected function AddRow($aRow)
 	{
 		$aData = array();
 		foreach ($this->aCSVHeaders as $sHeader) {
-			if (is_null($aRow[$sHeader]) && $this->AttributeIsNullified($sHeader)) {
+			if (strlen($aRow[$sHeader])==0 && $this->AttributeIsNullified($sHeader)) {
```
Molkobain, line 597:
```suggestion
			if (strlen($aRow[$sHeader] ?? '') === 0 && $this->AttributeIsNullified($sHeader)) {
```

Otherwise in case a null value (as it seem it can be possible) it will trigger a warning with PHP 8.

Molkobain:
> I'm only concerned on how to sync a NULL value into a nullable DECIMAL field in iTop. I think that is still not possible, even with these additions?
> 
> Edit: This is the correct link to the documentation: https://www.itophub.io/wiki/page?id=extensions:itop-data-collector-base#reset_or_untouch

Could we add a unit test on that to check and to ensure that we keep the intended behavior?

Hipska:
Hi @odain-cbd, what is the status / progress on this PR?

core/collector.class.inc.php
odain-cbd: thank you. correction is pushed like recommended

odain-cbd:
> > I'm only concerned on how to sync a NULL value into a nullable DECIMAL field in iTop. I think that is still not possible, even with these additions?
> > Edit: This is the correct link to the documentation: https://www.itophub.io/wiki/page?id=extensions:itop-data-collector-base#reset_or_untouch
> 
> Could we add a unit test on that to check and to ensure that we keep the intended behavior?

there is a test that covers nullify feature.

odain-cbd:
> Hi @odain-cbd, what is the status / progress on this PR?

Hi Hipska,
 
current PR was on hold on my side. still waiting for reviewers to approve.

Hipska:
That test doesn't actually test if the value on iTop becomes NULL after synchro. It only tests if resulting CSV contains `<NULL>`.

So question remains for very long time; How to synchronise a DECIMAL field to have the resulting value in iTop be NULL?

IMHO it has never been able to achieve this.

core/collector.class.inc.php
```diff
@@ -594,7 +594,7 @@ protected function AddRow($aRow)
 	{
 		$aData = array();
 		foreach ($this->aCSVHeaders as $sHeader) {
-			if (is_null($aRow[$sHeader]) && $this->AttributeIsNullified($sHeader)) {
+			if (strlen($aRow[$sHeader] ?? '')===0 && $this->AttributeIsNullified($sHeader)) {
```
Hipska, line 597: Empty string is not the same as `null`.

Hipska: No, The whole thing is to be able to set a field's value as `NULL` in the SQL database, not reset, not ignore!

So it seems it will still not work and Combodo still doesn't understand the needs.. 😢

PR #53, opened 2024-06-13
Fix N°7338 to actually use "itop_login"

Hipska:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | N°7338

## Symptom

Since #43, the documentation said it would use `itop_login` as fallback if no `synchro_user` was set. But this is wrong and it does not set any `synchro_user`.

## Reproduction procedure

1. With any collector
2. Do not set a value for `synchro_user`
3. Run `exec.php --configure_only`
4. Observe on iTop the **User** field of the _Synchro Data Source_ to be empty.

## Cause

The problem exists because the `Parameters::Get` method only checks if the key exists and then returns it, but since `<synchro_user></synchro_user>` is present in the `params.distrib.xml` file, it never tries to get the value for `itop_login`.

## Proposed solution
See code

### DIFF
--- a/core/orchestrator.class.inc.php
+++ b/core/orchestrator.class.inc.php
@@ -175,7 +175,7 @@ class Orchestrator
 				Utils::Log(LOG_ERR, "Unable to find the contact with email = '$sEmailToNotify'. No contact to notify will be defined.");
 			}
 		}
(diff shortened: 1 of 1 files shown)


jf-cbd:
Accepted in functional review

jf-cbd:
Merged, thanks for your contribution :D

PR #54, opened 2024-06-13
Fix default date_format

Hipska:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | #45

## Symptom

Some collectors didn't seem to synchronise a date(time) field anymore (since 1.4.0)

## Reproduction procedure

1. With teemip-ip-discovery-collector 3.1.3
4. Run a collection + synchro
5. Then observe in iTop that the **Last discovery date** isn't updated.

## Cause

itop-data-collector-base 1.4.0 introduced a new setting with a default value which wasn't the same as the previous behaviour.

## Proposed solution

Use the previous behaviour as the default value for the `date_format` setting. The other change provides useful debug data even when no errors detected on the import, like:
> 0: Wrong format for DATETIME column 3: ‘2024-06-13 14:05:09’ does not match the expected format: ‘Y-m-d’ (column skipped)

### DIFF
--- a/conf/params.distrib.xml
+++ b/conf/params.distrib.xml
@@ -107,5 +107,5 @@
   <!-- iTop user set as allowed to run synchronization. It is highly recommended to use the same as itop_login. By default, it is the same as itop_login -->
   <synchro_user></synchro_user>
   <!-- date format in collected data -->
-  <date_format>Y-m-d</date_format>
+  <date_format>Y-m-d H:i:s</date_format>
 </parameters>
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -692,7 +692,7 @@ abstract class Collector
 				'output' => $sOutput,
 				'csvdata' => file_get_contents($sDataFile),
 				'charset' => $this->GetCharset(),
-                'date_format' => Utils::GetConfigurationValue('date_format', 'Y-m-d')
+                'date_format' => Utils::GetConfigurationValue('date_format', 'Y-m-d H:i:s')
 			);
 
 			$sLoginform = Utils::GetLoginMode();
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -708,6 +708,8 @@ abstract class Collector
 				Utils::Log(LOG_ERR, $sTrimmedOutput);
 
 				return false;
+			} else {
+				Utils::Log(LOG_DEBUG, $sTrimmedOutput);
 			}
 		}
 


Blackwhite-Edward:
I Use  itop-data-collector-base master has a problem to collector vsphere ,
image
image
 there has a problem code to me when i run the scripts 
network@netbox:/data/vSphereCollector$ php exec.php --config_file=collectors/extensions/params.172.25.180.245.xml 

**error code**
PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in /data/vSphereCollector/collectors/src/vSphereModelCollector.class.inc.php:47

Hipska:
@Blackwhite-Edward What does this has to do with my pull request? It also seems like it has noting to do with this repository at all, since the error comes from a file not present here..

Blackwhite-Edward:
sry, i take the problem at a worng place

Hipska:
Should we really need to wait a year for a patch release to fix a backward incompatibility introduced in a minor version?

jf-cbd:
Hey @Hipska,
Thanks for your proposition.
Changing the default format of the collector base could create issues with other existing collectors (e.g. LDAP one), so it seems better not to change it current behavior.
@xtophe38 should release soon a new version of teemip-ip-discovery-collector that will include your [PR](https://github.com/TeemIp/teemip-ip-discovery-collector/pull/9) - thanks for it ! - so it will work in 3.1.4 with current itop-data-collector-base version.
Until that, you can override the date_format by defining it in conf/params.local.xml

jf-cbd:
As you know, different collectors can use different formats. The choice has been made to use "d/m/Y" format. Before the PR #45 , it was a problem with some other collector.
There is no universal format, that's why collectors can come with their own formats.
With your change in teemip-ip-discovery-collector will automatically override the change of itop-data-collector-base.

Hipska:
The universal format is the one iTop API's use as the default. The collector-base should reflect that. I highly doubt the "d/m/Y" format was chosen deliberately as it broke collectors depending on it.

jf-cbd:
I can re-submit the idea of changing it - it doesn't seem to be a bad idea. But yes, this is the choice that was made.

Hipska:
Yes please do. It is a very weird choice and as expressed multiple times, it broke existing collectors.

My PR also includes additional debug output for example when there is a mismatch in the date format, so users can find the problem easier..

jf-cbd:
I will.

The debug part is a good idea, for sure.

That said, no need to write some part of your [message](https://github.com/Combodo/itop-data-collector-base/pull/54#issuecomment-2684598589) in capital letters, I read better lowercase letters and words 😉

Hipska:
Perfect! It just seemed like you didn't see/read me mentioning that a previous change changed the default format. Just wanted to make sure you get that, as you implied that I was doing a change, while in fact, I'm only restoring that previous change.

jf-cbd:
We read it carefully with @xtophe38, and we talked about different changes that were made and proposed in different PRs on. We also did read internal tickets and choice that were taken in it. Based on that, I wrote this [message](https://github.com/Combodo/itop-data-collector-base/pull/54#issuecomment-2684594236).
I will submit your proposition - that make sense according to me - and keep you updated about the decision that will be made.

jf-cbd:
Hello, we agree with you : the format `Y-m-d H:i:s` should be used as default in itop-data-collector-base, especially considering other collectors and the iTop default synchro format.

Could you please change the default date format in `collector.class.inc.php` (line 695) ? Even if it's overwritten by your change, it's better to be consistent.

jf-cbd:
Thanks for your contribution !

PR #55, opened 2025-01-02
N°7783 - Relative paths cannot beused when the collector is not launched from its home directory

xtophe38:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=7783

## Symptom (bug) / Objective (enhancement)
The json collector included in itop-data-collector-base may get its source information from a json file located on the server hosting the collector. In such case, the parameter jsonfile defines the name of the file to be used by the collector together with its path to access it. The path may be absolute or relative to the directory where the collector is deployed (let's call it <collector>).

When a relative path is used to reference the json file (for instance data/myjsonfile.json) and when the collector is not launched from the directory <collector>, the json collector cannot locate the json file, logs the following error and stops.

Error:
PHP Warning:  file_get_contents(data/myjsonfile.json): Failed to open stream: No such file or directory in /my-root-dir/my-collector/core/jsoncollector.class.inc.php on line xyz
...
[2025-01-02 13:31:26]	[Error]	[MyJsonCollector] Failed to get JSON file: 
[2025-01-02 13:31:26]	[Error]	MyJsonCollector::Prepare() returned false

When can live with the bug by always specifying absolute paths in the configuration file. This implies that specific params.local.xml file must be provided for phpunit tests. The PR get rid of this requirement.

## Reproduction procedure (bug)
1. With itop-data-collector-base 1.4.0
2. With PHP 8.3.15
3. Deploy a collector base under a working directory that we'll name <my-collector>
4. Create a json collector. Don't specify any jsonurl parameter but specify a jsonfile one using a relative path like 'data/myjsonfile'.
5. Run the collector from <my-collector>, everything will work fine.
6. If you run it from another directory, the collector will break.

## Cause (bug)
The json collector doesn't handle the case where the source file may be defined with a relative path. See line 153 of core/jsoncollector.class.inc.php:

## Proposed solution (bug and enhancement)
Add a test after the first read of the jsonfile. If it fails, retry with the APPROOT constant prepended to it.

- [x] I have tested all changes I made on an iTop instance
- [x] I have added a unit test, otherwise I have explained why I couldn't

### DIFF
--- a/core/jsoncollector.class.inc.php
+++ b/core/jsoncollector.class.inc.php
@@ -150,7 +150,11 @@ abstract class JsonCollector extends Collector
 			Utils::Log(LOG_DEBUG, 'Source sFileJson: '.$this->sFileJson);
 			Utils::Log(LOG_INFO, 'Synchro URL (target): '.Utils::GetConfigurationValue('itop_url', array()));
 		} else {
-			$this->sFileJson = file_get_contents($this->sFilePath);
+            $this->sFileJson = file_get_contents($this->sFilePath);
+            if ($this->sFileJson === false) {
+                $this->sFilePath = APPROOT.$this->sFilePath;
+                $this->sFileJson = file_get_contents($this->sFilePath);
+            }
 			Utils::Log(LOG_DEBUG, 'Source sFileJson: '.$this->sFileJson);
 			Utils::Log(LOG_INFO, 'Synchro  URL (target): '.Utils::GetConfigurationValue('itop_url', array()));
 		}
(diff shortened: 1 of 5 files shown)


accognet (approved):

xtophe38:
Note: as you can see, the 2 provider functions of the JsonCollectorTest class have been made static. This is to comply with the requirements of the latest phpunit versions. I'm using v 10.5.29.

odain-cbd (approved):

PR #56, opened 2025-02-05
N°8162 - Issue when synchronizing datasource with read-only notify_contact_id_archive_flag

odain-cbd:
### **Purpose**
Current PR aims at gathering all special attributes at the same place. These attributes should not be taken into account to deal with datasource computation/synchronization. 

### **Symptom**
Datasource synchronization may fail with below error due to a regression (PR 32/ N°6115):
`[Error] Failed to update the SynchroDataSource 'XXX' (2). Reason: Error: notify_contact_id_archive_flag: Attempting to set the value on the read-only attribute SynchroDataSource::notify_contact_id_archive_flag (100)`

### **Corollary**
Datasource computation has been enhanced and related test coverage completed...

### DIFF
--- a/core/collector.class.inc.php
+++ b/core/collector.class.inc.php
@@ -979,7 +979,7 @@ abstract class Collector
 		foreach ($aDS1 as $sKey => $value) {
 			if (in_array($sKey, Collector::READONLY_FIELDS)){
 				// Ignore all read-only attributes
-				break;
+				continue;
 			}
 
 			switch ($sKey) {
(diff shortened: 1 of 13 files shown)


core/collector.class.inc.php
```diff
@@ -952,19 +977,14 @@ protected function FindAttr($sAttCode, $aExpectedAttrDef)
...
-					break;
+			if (in_array($sKey, Collector::READONLY_FIELDS)){
+				// Ignore all read-only attributes
+				continue;
+			}
 
+			switch ($sKey) {
 				case 'attribute_list':
-					foreach ($value as $sKey => $aDef) {
+					foreach ($value as $sKey2 => $aDef) {
```
Hipska, line 987: Since it is unused, better do this
```suggestion
					foreach ($value as $aDef) {
```

Hipska (approved):

core/collector.class.inc.php
```diff
@@ -41,6 +41,15 @@ abstract class Collector
 	 * @var string TABLENAME_PATTERN used to validate data synchro table name
 	 */
 	const TABLENAME_PATTERN = '/^[A-Za-z0-9_]*$/';
+	const READONLY_FIELDS = [
```
rquetiez, line 44: 1) Such a list depends on the datamodel: depending on class properties, the list of magical fields (related to obsolescence or archive) may vary.
The class "User" is not often subject to change, but The class "Contact" can be customized easily.
2) It is weird that you did not encounter the case with obsolescence_date and archive_date.

core/collector.class.inc.php
accognet: You can also remove this $sKey variable (as on line 987), it is ambiguous.

PR #60, opened 2025-09-10
Add detection of iTop API versions 1.3 and 1.4.

jbostoen:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | No
| Type of change?                                               | Enhancement

## Symptom (bug) / Objective (enhancement)

Enhancement: Ability to detect iTop API versions 1.3 and 1.4 ( https://www.itophub.io/wiki/page?id=latest:advancedtopics:rest_json#changes_history )

It also changes the logic to try most recent versions of the API first.

## Proposed solution (bug and enhancement)

Just add the missing values.

- [x] I have tested all changes I made on an iTop instance

## Checklist of things to do before PR is ready to merge
<!--
Things that needs to be done in the PR before it can be considered as ready to be merged

I'm not sure if Combodo internally has a version 2.0 of the API or why it's here already.
If needed, I'll add it back. Otherwise, it's just another API call that doesn't make any sense yet.

If no version is found, it will throw an exception now.

### DIFF
--- a/core/restclient.class.inc.php
+++ b/core/restclient.class.inc.php
@@ -129,21 +129,29 @@ class RestClient
 		return $aResults;
 	}
 
-	public static function GetNewestKnownVersion()
-	{
-		$sNewestVersion = '1.0';
+
+	/**
+	 * Gets the iTop API version in use. It tries all the known versions, starting from the newest one.
+	 *
+	 * @return void
+	 * @throws Exception If no supported version is found.
+	 */
+	public static function GetNewestKnownVersion() {
+
 		$oC = new RestClient();
-		$aKnownVersions = array('1.0', '1.1', '1.2', '2.0');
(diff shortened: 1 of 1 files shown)


core/restclient.class.inc.php
```diff
@@ -129,21 +129,29 @@ protected static function ExecOperation($aOperation, $sVersion = '1.0')
...
 	}
 
-	public static function GetNewestKnownVersion()
-	{
-		$sNewestVersion = '1.0';
+
+	/**
+	 * Gets the iTop API version in use. It tries all the known versions, starting from the newest one.
+	 *
+	 * @return void
```
Hipska, line 136: This method is not returning a void

Hipska (changes requested):

PR #67, opened 2026-01-09
fix(LookupTable): Also skip search if field is null

Hipska:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | N/A

## Symptom

When having nullified fields as part of the lookup, you will get a lot of these warning messages:

> ```
> [Warning]	No mapping found with key: '<NULL>_<NULL>', 'fieldname' will be set to zero.
> ```

## Reproduction procedure

1. Create a .raw.csv file where fields used in the lookup have `<NULL>` as value.
2. Run synchro.

## Cause

The method `Lookup` only skips if the fields are empty.

## Proposed solution

Also take the `NULL_VALUE` in consideration to skip lookup.


jf-cbd:
Hello, thank you for your PR. This warning seems to be make sense, it indicates that in case of value not found in the mapping table, and 0 value will be set instead. Why do you want to delete it ?

Hipska:
The warning does not happen when the value is an empty string, so it does not make much sense to have the warning when the value is null. It was forgotten to be added when nullified field feature was created.

